Issue #3, opened 2019-04-19
Verify Proofs against Spec


ethanfrey:
We have a proof that the preimage of the leaf node is the given bytes and that cannot be faked. However, as we allow flexibility in defining how key, value is transformed to the lead preimage, there is space for maliability. Or claiming a different subset of that preimage is actually the key and value.

To compact that we define a ProofSpec which is a list of requirements that all proofs from a given backend will meet. So, an Iavl Proof must match one spec, while a merk-js proof matches another spec.

These specs are also serialize able protobuf objects, not special code paths, to allow them to be registered and stored as configuration. And New chains to be added without adding client side code. 

Please add the spec struct and logic to check it

ethanfrey:
Update: mostly closed by #9 
Lacking spec enforcement code for js client

ethanfrey:
Fully closed via #13

PR #9, opened 2019-04-23
Fix proof specs

ethanfrey:
Addresses #3 

Add a specification for proof formats and enforce them on the ExistenceProof. Especially this is done to avoid maleability in selecting the key and value from the leaf node preimage.

- [x] Define ProofSpec format
- [x] Add checks in go code and test
- [x] Check ProofSpec with testvectors
- [ ] Support proof spec in js code


ethanfrey:
Merging without js support... will add that shortly

PR #12, opened 2019-07-30
WIP: Solidity Implementation

mossid:


ethanfrey:
Wow, interesting work.

I will cut a v0.2.0 with the current protobuf spec (which you used).

Can you add some commands in the readme I can use to run the test code?
I am happy to add this to the (to be added) CI as well, so I can check it faster

cwgoes:
bump @mossid @ethanfrey - this is very interesting (glad to review if you like).

ethanfrey:
Thanks for the bump, I didn't notice this was ready

ethanfrey:
@mossid Code looks good and good reuse of test cases. I need to dig in deeper, but a few meta-comments (on layout not solidity code)

But first, can you remove WIP if this is ready for review?

ethanfrey:
I don't like adding the ethereum dependency to go.mod/go.sum for the pure go implementation. It seems to be impossible to separate out release/test dependencies with go mod (https://github.com/golang/go/issues/26913), but maybe you can do what I did for the other one, base your `go.mod` in confio/ics23/sol and keep the deps separate (they are just for testing, right?)

* Huge go.mod in root now: https://github.com/confio/ics23/pull/12/files#diff-37aff102a57d3d7b797f152915a6dc16
* confio/ics23/go now depends on go-ethereum: https://github.com/confio/ics23/pull/12/files#diff-9998b2a22798a81c1fd814c2d90ad3da

I'm actually not sure about licenses since go-ethereum is GPL, so definitely do not want to import it in Apache license production code. I'm fine with it in the test code for solidity, but the import should be nowhere else.

ethanfrey:
I can see the solidity code in `sol/ics23.sol`, and that it is converted to go code for testing in ` sol/ics23_contract.go`, but... where is the step to do that transform?

Can you add a Makefile (or at least a README with the commands)? Nice to help for reproducing. Also if I change a line of solidity, I want to re-run the tests.

ethanfrey:
Please add a job to `.circleci/config.yml` to run the tests in the solidity folder. Ideally this also compiles the solidity to the ics23_contract.go file, so we ensure that we really are approving this exact solidity code.

ethanfrey:
I'll take a deeper code dive likely Monday. It would be great if you could do some of this cleanup first, so I just test the solidity implementation. 

I kind of feel that there is better support for testing solidity in javascript. I have used https://github.com/xf00f/web3x before as it was typescript and had some nice compilation steps and all, but I know there is a [wide array of tools](https://media.consensys.net/an-definitive-list-of-ethereum-developer-tools-2159ce865974). Not to say the go-ethereum tests are not valid, just that it might be easier to do the code-test-debug cycle with another framework.

In the end, the only real artifact of this is the *.sol files, not the test harness, so if it is documented and runs on the CI, I am fine with it. Just a suggestion if it is easier.

mossid:
> I don't like adding the ethereum dependency to go.mod/go.sum for the pure go implementation. It seems to be impossible to separate out release/test dependencies with go mod ([golang/go#26913](https://github.com/golang/go/issues/26913)), but maybe you can do what I did for the other one, base your `go.mod` in confio/ics23/sol and keep the deps separate (they are just for testing, right?)
> 
> * Huge go.mod in root now: https://github.com/confio/ics23/pull/12/files#diff-37aff102a57d3d7b797f152915a6dc16
> * confio/ics23/go now depends on go-ethereum: https://github.com/confio/ics23/pull/12/files#diff-9998b2a22798a81c1fd814c2d90ad3da
> 
> I'm actually not sure about licenses since go-ethereum is GPL, so definitely do not want to import it in Apache license production code. I'm fine with it in the test code for solidity, but the import should be nowhere else.

Agreed, the ethereum imported code is used for test contract deployment and hash functions. It certainly can be separated and remove the dependency on root go.mod.

> I can see the solidity code in `sol/ics23.sol`, and that it is converted to go code for testing in ` sol/ics23_contract.go`, but... where is the step to do that transform?
> 
> Can you add a Makefile (or at least a README with the commands)? Nice to help for reproducing. Also if I change a line of solidity, I want to re-run the tests.

I used [Native Dapps](https://github.com/ethereum/go-ethereum/wiki/Native-DApps:-Go-bindings-to-Ethereum-contracts), will add a makefile.



> Please add a job to `.circleci/config.yml` to run the tests in the solidity folder. Ideally this also compiles the solidity to the ics23_contract.go file, so we ensure that we really are approving this exact solidity code.

:+1:



> I'll take a deeper code dive likely Monday. It would be great if you could do some of this cleanup first, so I just test the solidity implementation.
> 
> I kind of feel that there is better support for testing solidity in javascript. I have used https://github.com/xf00f/web3x before as it was typescript and had some nice compilation steps and all, but I know there is a [wide array of tools](https://media.consensys.net/an-definitive-list-of-ethereum-developer-tools-2159ce865974). Not to say the go-ethereum tests are not valid, just that it might be easier to do the code-test-debug cycle with another framework.
> 
> In the end, the only real artifact of this is the *.sol files, not the test harness, so if it is documented and runs on the CI, I am fine with it. Just a suggestion if it is easier.

I am not that familiar with Javascript, I think we can use the current codebase and switch it to javascript later(I agree that javascript has better support on solidity testing and we should switch to it)

ethanfrey:
@mossid The comment sounds good. Did you have time to work on this? (I know you've been busy with DevCon and IBC work)

mossid:
Working on this PR for future ibc-eth integration

ethanfrey:
> Working on this PR for future ibc-eth integration

@mossid I think the types changed since you worked on this. They are quite stable now. You may want to restart with proper types?

ethanfrey:
I would check with the IBC team from IG maybe they have some ideas I haven't thought about

AdityaSripal:
Hello @lightyear15 , thank you for the detailed writeup. As you say, we do not currently use `GetProofSpecs` interface externally.

We do use the `ProofSpecs` in `ValidateSelfClient` during the connection handshake: https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/keeper/keeper.go#L301

We discussed keeping the interface public so that we can use it when we relax the `ValidateSelfClient` function to also accept WASM tendermint clients. However, this would still require casting to a specific client type to perform our other checks so keeping `ProofSpecs` in the interface is not so useful.

We are planning to greatly reduce the interface for `ClientState` and put all responsibility on individual client implementations.
This is discussed in the WASM spec PR. That discussion kept `GetProofSpecs` as part of the interface as mentioned above, but given the necessity of casting to particular client types, it is going to be removed as well.
https://github.com/cosmos/ibc/pull/571#discussion_r654342052


For now, you can ignore the `GetProofSpecs` method and use your own verification method for Patricia tries as you say. Note this will require you to implement your own proof verification functions rather than using the generic merkle proof verifier provided in ics23.

lightyear15:
@adizere, @andynog,
you might want to join in and have some fun :smile:

ethanfrey:
Closing in favour of #58

PR #20, opened 2019-08-12
Rust client

### LINKED ISSUE #6: Add rust verification code
Add code in rust to validate confio/proofs objects. Adding support for the third major language in use in  blockchain projects, and extending compatibility.

ethanfrey:
Closes #6 

- [x] Compile and check protobuf
- [x] Clean protoc tooling (not needed to import)
- [x] Implement and test validation code
- [x] CI support
- [x] Support all test vectors
- [x] Enforce Spec


rust/codegen/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+fn main() {
+	let args: Vec<_> = env::args().collect();
+	let mut root = "../..";
+	if args.len() > 1 {
+		root = &args[1];
+	}
+
+	protoc_rust::run(protoc_rust::Args {
+	    out_dir: &format!("{}{}", root, "/rust/src"),
+	    input: &[&format!("{}{}", root, "/proofs.proto")],
```
spacejam: For windows compatibility, you might want to use `env::var("OUT_DIR")` (see [the codegen section of the book for more on that](https://doc.rust-lang.org/cargo/reference/build-scripts.html#case-study-code-generation)) to get the root and [PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) to handle path separators properly, and then the `to_str` method to format it as an optional `&str` (it's an `Option` because windows paths are utf-16 and `str` is guaranteed to be valid utf-8, and maybe something got messed up.)

rust/codegen/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+	    includes: &[root],
+	    customize: Customize {
+	      ..Default::default()
+	    },
```
spacejam: Depending on your desired level of clarity, this could also be written as 
```
    includes: &[root],
    customize: Default::default(),
}
```
or
```
    includes: &[root],
    ..Default::default(),
}
```

but it's not necessarily bad to be explicit here that this is the type being used

rust/src/helpers.rs
```diff
@@ -0,0 +1,4 @@
+extern crate failure;
+
+pub type Result<T> = ::std::result::Result<T, failure::Error>;
+pub type Hash = ::std::vec::Vec<u8>;
```
spacejam: while the leading colons on `::std` are not incorrect, they are not necessary either, and most of the time people omit them

rust/src/ics23.rs
```diff
@@ -0,0 +1,128 @@
...
+        verify_test_vector("../testdata/tendermint/existence3.json", &spec)
+    }
+
+    #[test]
+    fn test_vector_tendermint4() -> Result<()> {
+        let spec = tendermint_spec();
+        verify_test_vector("../testdata/tendermint/existence4.json", &spec)
+    }
+
+}
```
spacejam: there are a few minor line spacing things in this file that rustfmt can clean up for you automatically. Usually folks avoid multiple blank lines line at the top of the file, or blank lines at the end of a block like on line 127

rust/src/ics23.rs
```diff
@@ -0,0 +1,128 @@
...
+use crate::verify::{CommitmentRoot, verify_existence};
+
+pub fn verify_membership(proof: &proofs::CommitmentProof, spec: &proofs::ProofSpec, root: CommitmentRoot, key: &[u8], value: &[u8]) -> bool {
+    match &proof.proof {
+        Some(proofs::CommitmentProof_oneof_proof::exist(ex)) => {
+            let valid = verify_existence(&ex, spec, root, key, value);
+            valid.is_ok() && valid.unwrap()
+        }
+        _ => false,
+    }
```
spacejam: when matching on a single variant of an enum, you can avoid nesting by writing it as:
```
if let Some(proofs::CommitmentProof_oneof_proof::exist(ex))) = &proof.proof {
    let valid = verify_existence(&ex, spec, root, key, value);
    valid.is_ok() && valid.unwrap()
} else {
    false
}
```

rust/src/ics23.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    extern crate protobuf;
+    extern crate serde;
+    extern crate serde_json;
```
spacejam: in general, `extern crate` is not necessary when using rust 2018, with a [couple rare exceptions](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#an-exception)

rust/codegen/src/main.rs
ethanfrey: I kind of explicitly don't support windows for dev environments (you can run the code, but not necessarily contribute). Good tip here, but I'll leave out windows

rust/codegen/src/main.rs
ethanfrey: Ah, that was cut and paste, interesting to understand what it is doing

rust/src/ics23.rs
ethanfrey: Ah... I only saw this after asking on zulip :man_facepalming:

rust/src/helpers.rs
ethanfrey: cool, the link as to the why (rust 2018) was good

ethanfrey:
Thanks for the review

PR #24, opened 2019-08-22
Add support for Batch proofs

### LINKED ISSUE #19: Add efficient storage format for BatchProofs
We should be able to represent a set of `ExistenceProofs` and `NonExistenceProofs` as a Batch, and add some encoding to require much less space to store those serialized proofs than naively concatenating them.

This allows us to send multiple messages over IBC at once more efficiently.

ethanfrey:
Closes #19 

- [x] Update Protobuf defs for better proof and compressed proof
- [x] Extend MemberShip, NonMembership for to work with batch proofs
  - [x] Go
  - [x] Rust
  - [x] Typescript
- [x] Add BatchMembership, BatchNonMembership verification
  - [x] Go
  - [x] Rust
  - [x] Typescript
- [x] Add Compression/Decompression for BatchProofs (and integrate with verify calls)
  - [x] Go
  - [x] Rust
  - [x] Typescript


ethanfrey:
Okay, this is done in go. I can maybe add more testvectors from a child package, specifically for a batch proof from one tree.

I am happy for feedback on the algorithm implemented in go, before implementing it in two more languages.

go/ics23.go
```diff
@@ -44,11 +50,84 @@ func VerifyMembership(spec *ProofSpec, root CommitmentRoot, proof *CommitmentPro
...
+		if !valid {
+			return false
+		}
+	}
+	return true
+}
+
+// BatchVerifyNonMembership will ensure all items are also proven to not be in the Commitment by the CommitmentProof
+// (which should be a BatchProof, unless there is one item, when a NonExistenceProof may work)
+func BatchVerifyNonMembership(spec *ProofSpec, root CommitmentRoot, proof *CommitmentProof, keys [][]byte) bool {
```
cwgoes: Should there be another function or a flag to indicate contiguous keys for which a range proof could be used (if supported by the underlying tree)?

go/ics23.go
```diff
@@ -44,11 +50,84 @@ func VerifyMembership(spec *ProofSpec, root CommitmentRoot, proof *CommitmentPro
...
 	return err == nil
 }
+
+// BatchVerifyMembership will ensure all items are also proven by the CommitmentProof (which should be a BatchProof,
+// unless there is one item, when a ExistenceProof may work)
+func BatchVerifyMembership(spec *ProofSpec, root CommitmentRoot, proof *CommitmentProof, items map[string][]byte) bool {
+	// decompress it before running code (no-op if not compressed) - once for batch
+	proof = Decompress(proof)
+	for k, v := range items {
+		valid := VerifyMembership(spec, root, proof, []byte(k), v)
```
cwgoes: Should there be another function or a flag to indicate contiguous keys for which some proof data could be shared?

ethanfrey: Well the proof data is shared automatically when you compress it. For any batch proof, whether it is range or not. Basically, if an inner node is present N times, it will be shrinked down to around (1 + N/16).

What it doesn't capture explicitly is the proof that all these N keys are continuous. There is the code to check `IsLeftNeighbor(spec, left, right)` which can be fed these items in batch and reconstruct they are a range.

go/ics23.go
ethanfrey: I would leave it as is for now (fits just the ics23 spec), but let's use this as some feedback on the spec, and then I can enhance these functions to meet the final spec.

One point is that there is no such function as `VerifyMembership(proof, key, value)`. It needs to know the context of the type of tree. You could rewrite it to example:

`VerifyMembership: (spec) => (proof, key, value) => bool`

or anything else, but capturing some spec (definition of a particular merkle format) is critical, especially for non-membership proofs.

ethanfrey: The idea of range/batch/neighbor calculations can definitely be expanded.

You may note that I expose non-membership proofs for stores that only themselves have a membership proof (ics23-tendermint). If the underlying store is ordered by key, we can produce all kinds of proofs, with a bunch of processing (client-side?) on the native proof format.

I'd be happy for more discussion on the ics23 spec, informed by what is built here (and what seems easy/hard to add)

PR #28, opened 2020-03-16
Update proto spec

ethanfrey:
This addresses concerns raised by @cwgoes here: https://github.com/cosmos/ics/issues/282#issuecomment-589690471

Notably, the spec now fixes the hash algorithm for inner nodes and optionally allows setting a min and max depth for the trees (min/max number of inner nodes).


ethanfrey:
> ACK, although do you want to add tests for the spec mis-matching (hash / min depth / max depth)?

There is a test for hash. I will add some for min/max depth. As well as the Rust and JS implementations.

> I believe that this addresses all the specifically ICS23-related concerns in cosmos/ics#282 (comment); we still need to check the CommitStore and sub-store logic though (in any case).

This is changes in the SDK. I can help with those as well, but that is not in this repo. I would demo how to use `ics23/tendermint` to make proofs for the CommitStore (but we need to change it to use merkle proofs, not just amino-serialize the roots of all sub-stores and hash it)

ethanfrey:
@cwgoes I extended the tests for go to enforce min/max depth and made the same test cases for both TypeScript and Rust and updated validation logic in both.

I believe this is completed. Happy for one more review from you, and will merge and tag upon approval

PR #30, opened 2020-05-22
Add Calculate method to CommitmentProof

AdityaSripal:


go/proof.go
```diff
@@ -40,6 +40,34 @@ var TendermintSpec = &ProofSpec{
...
+// by type switching and calculating root based on proof type
+// NOTE: Calculate will return the first calculated root in the proof,
+// you must validate that all other embedded ExistenceProofs commit to the same root.
+// This can be done with the Verify method
+func (p *CommitmentProof) Calculate() (CommitmentRoot, error) {
+	switch v := p.Proof.(type) {
+	case *CommitmentProof_Exist:
+		return v.Exist.Calculate()
+	case *CommitmentProof_Nonexist:
+		return v.Nonexist.Left.Calculate()
```
AdityaSripal, line 53: Note left may be nil, have to check both sides

go/proof.go
```diff
@@ -40,6 +40,34 @@ var TendermintSpec = &ProofSpec{
...
+	case *CommitmentProof_Batch:
+		if len(v.Batch.GetEntries()) == 0 || v.Batch.GetEntries()[0] == nil {
+			return nil, errors.New("batch proof has empty entry")
+		}
+		bexist := v.Batch.GetEntries()[0].GetExist()
```
ethanfrey, line 58: You don't have to do that do you? Just recurse.

```go
return v.Batch.GetEntries()[0].Calculate()
```

go/proof.go
AdityaSripal: Returned error, since `ExistenceProof` will return error on an empty proof

AdityaSripal:
Added a calculate check on `TestVectors` that checks each proof calculates up to the expected root.

I couldn't add an equivalent `TestBatchVectors` since the proofs that are being batched together don't commit to the same root.

ethanfrey:
@AdityaSripal merged this and tagged it as v0.6.1 if you want to update the imports.

PR #31, opened 2020-05-28
Replace value with valuehash in existence proofs

AdityaSripal:


go/compress.go
```diff
@@ -77,10 +77,10 @@ func compressExist(exist *ExistenceProof, lookup *[]*InnerOp, registry map[strin
 		return nil
 	}
 	res := &CompressedExistenceProof{
-		Key:   exist.Key,
-		Value: exist.Value,
-		Leaf:  exist.Leaf,
-		Path:  make([]int32, len(exist.Path)),
+		Key:       exist.Key,
+		ValueHash: exist.ValueHash,
```
ethanfrey, line 81: Okay, so this is just the raw bytes includes in the leaf - we can also use KeyHash here, which would make it consistent.

I mean for IAVL I believe this is a NoHash for Key, so KeyHash = Key, but this would make it symetric

go/ops.go
```diff
@@ -15,24 +15,20 @@ import (
...
+// Apply will calculate the leaf hash given the key and value hash being proven
+func (op *LeafOp) Apply(key []byte, vhash []byte) ([]byte, error) {
 	if len(key) == 0 {
 		return nil, errors.New("Leaf op needs key")
 	}
-	if len(value) == 0 {
+	if len(vhash) == 0 {
 		return nil, errors.New("Leaf op needs value")
 	}
 	pkey, err := prepareLeafData(op.PrehashKey, op.Length, key)
```
ethanfrey, line 26: And if we use `khash` instead of `key`, this line disappears

go/proof.go
```diff
@@ -66,6 +59,23 @@ func (p *ExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []byte
 
 }
 
+// VerifyItem checks that the proof is a valid proof of a (key, value) for this root
+func (p *ExistenceProof) VerifyItem(spec *ProofSpec, root CommitmentRoot, key, value []byte) error {
+
+	if !bytes.Equal(key, p.Key) {
```
ethanfrey, line 65: Or rather, moves down here (only for VerifyItem)

go/proof.go
```diff
@@ -40,20 +40,13 @@ var TendermintSpec = &ProofSpec{
 	},
 }
 
-// Verify does all checks to ensure this proof proves this key, value -> root
-// and matches the spec.
-func (p *ExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []byte, value []byte) error {
+// Verify does all checks to ensure this proof is a valid existence proof for this root
+// NOTE: In order to prove a value, we must call VerifyItem
+func (p *ExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot) error {
```
ethanfrey, line 45: I understand that `Verify` will check the proof is valid (matches spec, and running the spec'd hash from leaf over all inner nodes will match the provided CommitmentRoot). But it doesn't say anything about what is being proven.

go/proof.go
```diff
@@ -66,6 +59,23 @@ func (p *ExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []byte
 
 }
 
+// VerifyItem checks that the proof is a valid proof of a (key, value) for this root
+func (p *ExistenceProof) VerifyItem(spec *ProofSpec, root CommitmentRoot, key, value []byte) error {
```
ethanfrey, line 63: VerifyItem then does both `Verify` and that the proof actually proves what we think it does.

go/proof_test.go
```diff
@@ -13,50 +13,50 @@ func TestExistenceProof(t *testing.T) {
...
+			expected: fromHex("7496fdaa49e3764d635f9f21e60cec0a75b8d7595a9a3bb013692bb45d14e326"),
 		},
 		// iavl leaf: start with 0, length 3
 		// inner prefix: !start with 0, length >= 4
 		"demonstrate maliability of leaf if we change leaf algorithm": {
 			proof: &ExistenceProof{
-				Key:   append([]byte{4}, []byte("food")...),
-				Value: append([]byte{16}, []byte("some longer text")...),
+				Key:       append([]byte{4}, []byte("food")...),
+				ValueHash: []byte("some longer text"),
```
ethanfrey, line 38: It would be best to keep the hashes all the same in these tests, so we see this doesn't change the proof format, just the metadata passed along with the proof. (Here I refer to leaf+inner as the proof, and key+value (maybe hashed?) as the metadata).

If you just use `ValueHash: append([]byte{16}, []byte("some longer text")...),` you should get the same expected result (as this would be using `PrehashKey` and `PrehashValue` as `NoHash` - https://github.com/confio/ics23/blob/master/proofs.proto#L109-L117)

go/proofs.pb.go
```diff
@@ -492,23 +490,21 @@ func _CommitmentProof_OneofSizer(msg proto.Message) (n int) {
...
 //hkey = prehashKey(key)
-//hvalue = prehashValue(value)
 //
 //Then combine the bytes, and hash it
-//output = hash(prefix || length(hkey) || hkey || length(hvalue) || hvalue)
+//output = hash(prefix || length(hkey) || hkey || valuehash)
 type LeafOp struct {
-	Hash         HashOp   `protobuf:"varint,1,opt,name=hash,proto3,enum=ics23.HashOp" json:"hash,omitempty"`
-	PrehashKey   HashOp   `protobuf:"varint,2,opt,name=prehash_key,json=prehashKey,proto3,enum=ics23.HashOp" json:"prehash_key,omitempty"`
-	PrehashValue HashOp   `protobuf:"varint,3,opt,name=prehash_value,json=prehashValue,proto3,enum=ics23.HashOp" json:"prehash_value,omitempty"`
```
ethanfrey, line 507: Why do you remove this? We actually need this info for validation.
I think you also refer to it on this line: https://github.com/confio/ics23/pull/31/files#diff-9f9b21196d67539dcb8d06f539269585R68

ethanfrey: You also know editing this file by hand is a no-no and you should edit proofs.proto in root, right?
(I accept it for you doing debuging, but this is needed to be fixed properly before any merge).

That also means we will need to update rust and js code with the new structs and also the generation code....

go/proof.go
cwgoes: Aye, should we even expose this function then? No IBC code should ever be calling `Verify` as defined like this.

cwgoes: Maybe just add `key, value []byte` as extra parameters? Do we need a separation here?

ethanfrey:
I think this got closed with other integration work. And there have been more feedback from the sdk team since then. Closing this. Let me know if there is anything here you want to pull out as an enhancement

Issue #35, opened 2020-08-27
Unable to use Go releases with go.mod


erikgrinaker:
Trying to use ics23 releases in IAVL, by putting this in `go.mod`:

```
require (
	github.com/confio/ics23 v0.6.3
)
```

Gives a `go mod tidy` error when importing like this:

```
import (
  ics23 "github.com/confio/ics23/go"
)
```

> github.com/confio/ics23/go is not in your go.mod file

Changing the `go.mod` module to include `/go` fails because the module does not have a v0.6.3 release, and changing the import to not use `/go` fails because there is no Go package at that path.

erikgrinaker:
CC @AdityaSripal

ethanfrey:
Yeah, I have no idea how to improve this. I try to keep language independent dirs under root, and go is the only one complaining. 

If you figure out a better way to handle go.mod, please make a pr and mention me. I'm more than happy to merge. (I spent some time with this when setting up the repo... that was go 1.12 or 1.13 so maybe things have changed)

zmanian:
My understanding of the issue is the release tag needs to have sub-directory in it.

so instead of `v0.6.3` it would be `go/v0.6.3`

erikgrinaker:
I tested this with a sandbox repo, there are two options:

1. Like @zmanian says, use an additional tag `go/v0.6.3` for each release to work around this.

2. Move `go/go.mod` to the root, and call the module `github.com/confio/ics23`. It must still be imported as `github.com/confio/ics23/go` in source files, but referenced as `github.com/confio/ics23` in the consumer's `go.mod`.

Personally I feel like 2 is the lesser of evils, especially since there's already a proxy `go.mod` in the root, but your call @ethanfrey.

ethanfrey:
I did try 2 before, and go mod made the 2 mod setup. This was not my idea, but I didn't spend a lot of time investigating it.

I tagged `go/v0.6.3` as per Zaki's request, I should add some info about this to the README I guess

zmanian:
I think this issue can be closed

erikgrinaker:
Yeah, I can use version tags in `go.mod` now, so I'm happy. Thanks guys!

PR #39, opened 2021-04-19
Add SHA-512/256 hash op and FIXED32_LITTLE length op

kostko:
See discussion in https://github.com/cosmos/ibc/issues/554.


ethanfrey (commented):
Thanks for this PR and nice to point out https://github.com/confio/ics23/pull/39/files#diff-946b8d2ee64b4b178e19710fbd2c831353ccef71968db74f051f482ca747afb4R167-R171

We aim to keep all 3 language implementations in sync, so this needs to be implemented for rust and typescript (in other folders).

If you can implement them, that would be great. If not, I will put this on my TODO list.

kostko:
Whoops, missed that, yeah I can add the other implementations.

kostko:
@ethanfrey Added the rest, Javascript is the most annoying.

js/src/ops.ts
```diff
@@ -152,3 +156,15 @@ function encodeVarintProto(n: number): Uint8Array {
   enc = [...enc, l];
   return new Uint8Array(enc);
 }
+
+function encodeFixed32LE(n: number): Uint8Array {
+  const enc = new Uint8Array(4);
+  let l = n;
+  for (let i = enc.length; i > 0; i--) {
+    /* tslint:disable */
+    enc[Math.abs(i - enc.length)] = l;
```
ethanfrey, line 165: I would do ` = (l % 256)`. That is quite clear what you want. I think this will happen automatically in JS, but it is harder to review and ensure correctness.

ethanfrey (approved):
Looks good. One minor comment.

If you do fix that, please run `cargo fmt` on the rust code as well. Then it is good to merge.

js/src/ops.ts
kostko: That wouldn't actually work correctly as the bits need to be truncated and using modulo would give you incorrect results. The most correct thing would actually be `l & 0xFF` but I avoided that because the linter didn't like it. But since we're shutting off the linter for mutation anyway I think we can add this to make it more clear.

kostko: Yeah makes sense, I prefer bitwise ops (e.g., `l & 0xFF`), but the linter doesn't like them for some reason.

kostko: Changed.

ethanfrey (approved):
Looks good. Thanks for your contribution.

ethanfrey:
I pushed 0.6.5 rust crate and npm package.

I tagged for go, but I got some errors trying to update local files. I will try again for that

kostko:
> I tagged for go, but I got some errors trying to update local files. I will try again for that.
> 
> There was a solution in #35 but I just get errors.

Seems to work for me with both `v0.6.5` and `v0.6.6` (since your module is in a subfolder you need to use e.g. `go/v0.6.6` as tags -- which you are). Note that you need to import this as `github.com/confio/ics23/go` so not sure why you also have a top-level `go.mod` file.

I would just remove that one (e.g., https://github.com/confio/ics23/blob/master/go.mod and corresponding `go.sum`) as it is confusing and unused.

ethanfrey:
> Note that you need to import this as github.com/confio/ics23/go so not sure why you also have a top-level go.mod file.

> I would just remove that one (e.g., https://github.com/confio/ics23/blob/master/go.mod and corresponding go.sum) as it is confusing and unused.

Yes, it is confusing, and Zaki helped with the Go tagging. I can gladly remove the top level go.mod. For some reason I thought it was needed.

ethanfrey:
BTW, if you need this in a build of the cosmos sdk, you need to manually update the dependency. And if you want it on the hub, maybe make a PR to bump the ics23 version, so it will be included in the next release

PR #41, opened 2021-06-11
support ics23 no-std

DaviRain-Su:
Support Ics23 rust no-std features.

### DIFF
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -1,5 +1,7 @@
 use std::collections::btree_map::BTreeMap as HashMap;
-use std::prelude::v1::*;
+
+#[cfg(not(feature = "std"))]
+use std::prelude::*;
 
 use crate::compress::{decompress, is_compressed};
 use crate::ics23;
@@ -204,14 +206,18 @@ pub fn tendermint_spec() -> ics23::ProofSpec {
 
 #[cfg(test)]
 mod tests {
+    extern crate std as _std;
     use super::*;
 
     use anyhow::{bail, ensure};
     use prost::Message;
     use serde::Deserialize;
-    use std::fs::File;
-    use std::io::prelude::*;
+    #[cfg(feature = "std")]
+    use _std::fs::File;
+    #[cfg(feature = "std")]
+    use _std::io::prelude::*;
     use std::vec::Vec;
+    use alloc::string::String;
 
     use crate::compress::compress;
     use crate::helpers::Result;
@@ -230,6 +236,7 @@ mod tests {
         pub value: Option<Vec<u8>>,
     }
 
+    #[cfg(feature = "std")]
     fn load_file(filename: &str) -> Result<(ics23::CommitmentProof, RefData)> {
         let mut file = File::open(filename)?;
         let mut contents = String::new();
@@ -252,6 +259,7 @@ mod tests {
         Ok((parsed, data))
     }
 
+    #[cfg(feature = "std")]
     fn verify_test_vector(filename: &str, spec: &ics23::ProofSpec) -> Result<()> {
         let (proof, data) = load_file(filename)?;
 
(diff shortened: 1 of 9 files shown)


ethanfrey:
Interesting and thank you for the contribution. I will look later.

In the meantime, could you update the CI scripts so they will run tests in no_std mode (`--no-defaut-features`, correct?) as well as the normal `std` mode?

DaviRain-Su:
> Interesting and thank you for the contribution. I will look later.
> 
> In the meantime, could you update the CI scripts so they will run tests in no_std mode (`--no-defaut-features`, correct?) as well as the normal `std` mode?

Ok ah, I'll see how to set it up.

ethanfrey:
Thank you.

ethanfrey:
I just checked out your branch and tried running no_std tests:

`cargo test --all --no-default-features`

This produces a number of errors. Most are either:

`cannot find function `decode` in crate `hex``
or
`vec![valid_inner.clone()]`

with suggestions for ` no_std_compat::prelude` imports.

ethanfrey:
Can you add these lines to the end of the circleCI job?

```yaml
      - run:
          name: Build all targets in no_std
          command: cargo build --all --no-default-features
       - run:
           name: Run all tests with no-std
           command: cargo test --all --no-default-features
```

DaviRain-Su:
> cargo test --all --no-default-features

It doesn't seem to work.

ethanfrey:
> > cargo test --all --no-default-features
> 
> It doesn't seem to work.

Exactly, the tests fail to compile (build works, but test code seems to not be no_std compatible).
Can you try fixing those, at least obvious issues. If something requires deeper changes to the codebase, I can take a look.

DaviRain-Su:
> > > cargo test --all --no-default-features
> > 
> > 
> > It doesn't seem to work.
> 
> Exactly, the tests fail to compile (build works, but test code seems to not be no_std compatible).
> Can you try fixing those, at least obvious issues. If something requires deeper changes to the codebase, I can take a look.

I solved this problem, see if you can pass it.

DaviRain-Su:
@ethanfrey Hi, Can you merge this PR?

rust/src/api.rs
```diff
@@ -230,6 +236,7 @@ mod tests {
         pub value: Option<Vec<u8>>,
     }
 
+    #[cfg(feature = "std")]
```
ethanfrey, line 239: Okay, so these tests are not run in no_std mode.

It gets the code to pass (and I guess I can merge it), but it would be nice to run the whole test suite with no-std

rust/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(dead_code)]
```
ethanfrey, line 2: funny smell here.
I guess some code is only exposed via std or no-std?
We could also just conditionally expose it.

ethanfrey (approved):
Sorry for the late review. I was on vacation the last two weeks.

Thank you for adding no-std to the CI and getting tests to pass. I created a follow up issue #42 that it would be great if you could look at sometime. But for now, I will merge your changes to unblock you. It would be nice if we could tackle at least some of #42 before I tag and push a new crate

rust/src/api.rs
DaviRain-Su: I think it is necessary to test here in the std environment, for `File` ,`io` and some other libraries are not available in `core` and `alloc`, but only in the standard library.
Relate tourtial: https://docs.rust-embedded.org/book/intro/no-std.html

rust/src/lib.rs
DaviRain-Su: Yes, compile by using `#![cfg_attr(not(feature = "std"), no_std)]` to conditionally select between `std` and `no_std`.

ethanfrey: I was commenting on `#![allow(dead_code)]`, which I removed in another PR

rust/src/api.rs
ethanfrey: Ah... you are very right.

I guess none of these work in no-std but it would be nice to have equivalent tests.
Maybe you can use [`include_bytes`](https://doc.rust-lang.org/std/macro.include_bytes.html), and then get rid of all the `load_file` stuff as well (and have tests work in both environments)

Issue #42, opened 2021-07-12
No-std cleanup


ethanfrey:
Follow up from #41 

- [ ] Enable the `api.rs` tests in no-std mode as well
- [ ] Remove `#![allow(dead_code)]` and `#![allow(unused_imports)]` from `lib.rs` if possible
- [ ] Fix clippy warnings and test in CI (currently `compress.rs:10` recommends using `matches!`... but maybe only as I use 1.51 locally and we run with 1.42 in CI... let's stick with 1.42 compatibility) 

Remove the need for `

ethanfrey:
I fixed some of this in #43 

I know little about no-std, but would be happy for someone to fix the rest.

PR #43, opened 2021-07-12
Fix linting issues

ethanfrey:
Part of #42 

- [x] Add `cargo fmt` and `cargo clippy` to CI
- [x] Fix fmt issues
- [x] Fix clippy issues
- [x] Remove #![allow(dead_code)] and #![allow(unused_imports)] from lib.rs


ethanfrey:
@DaviRain-Su can you please review this.

While trying to address some of the clippy warnings (and running 1.42 locally to get the same as the CI), I realized the `ics23.rs` file was manually editted and doesn't match what comes from `cargo protoc`, which will cause issues in the future keeping them in sync. I tried regenerating it with the script and got all kinds of errors, even in `std` mode.

Otherwise the rest was just making clippy happy basically

PR #44, opened 2021-07-19
Fix cargo protoc error

DaviRain-Su:
Fix : https://github.com/confio/ics23/pull/43
Relate issue: https://github.com/confio/ics23/issues/42

- [x] ensure ics23.rs can be regenerated via cargo protoc without breaking the build
- [x] Enable the api.rs tests in no-std mode as well

For `api.rs` also passes the test under `no_std` is problematic, due to the use of the `File` and `IO` libraries under the standard library. However, there is no core library in `core`, and `alloc`, so now it is possible to handle it in this way and only let it test under std.


DaviRain-Su:
@ethanfrey Hi, Can you review this PR?

DaviRain-Su:
Relate Rust no-std: https://justjjy.com/Rust-no-std

tzemanovic:
Thanks for updating this. I think you also need to bump prost version in `rust/Cargo.toml` and regenerate the proto files with the `rust/codegen`

rust/src/ics23.rs
```diff
@@ -36,8 +36,8 @@ pub struct ExistenceProof {
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct NonExistenceProof {
     /// TODO: remove this as unnecessary??? we prove a range
-    #[prost(bytes, tag = "1")]
-    pub key: alloc::vec::Vec<u8>,
+    #[prost(bytes = "vec", tag = "1")]
+    pub key: ::prost::alloc::vec::Vec<u8>,
```
ethanfrey, line 40: This looks better. I need to test it locally

DaviRain-Su: > This looks better. I need to test it locally

enen,  I tested it and it's OK.

DaviRain-Su:
> Thanks for updating this. I think you also need to bump prost version in `rust/Cargo.toml` and regenerate the proto files with the `rust/codegen`

Yes, I have been update cargo.toml.

ethanfrey:
Looks good.

I checked out the code and all tests pass.

Running `cargo protoc` and `cargo fmt` will regenerate the same code.

PR #47, opened 2021-07-30
Update prost version 0.7.0

DaviRain-Su:
Because the version of `prost` and `prost-type` in `ibc-rs` and `tendermint-rs` is still 0.7.0, I hope the `prost` used in `ics23` is downgraded to 0.7.0.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -53,7 +53,7 @@ name = "codegen"
 version = "0.1.0"
 dependencies = [
  "bytes",
- "prost",
+ "prost 0.7.0",
  "prost-build",
 ]
 
@@ -133,7 +133,7 @@ dependencies = [
  "anyhow",
  "bytes",
  "hex",
- "prost",
+ "prost 0.7.0",
  "ripemd160",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


ethanfrey:
Why downgrade?

Wouldn't it make more sense for those packages to upgrade?
(Unless there is some good reason they cannot upgrade, in which case I will downgrade)

Usually I update for compatibility with other crates

DaviRain-Su:
> Why downgrade?
> 
> Wouldn't it make more sense for those packages to upgrade?
> (Unless there is some good reason they cannot upgrade, in which case I will downgrade)
> 
> Usually I update for compatibility with other crates

Oh, I think it is temporarily reserved, well, generally speaking crate version is gradually upgraded, mainly I just tested in the detection of the other side depends on the library of the prost too much are 0.7 version. I think they will update the version later. This I also go to raise a issuse to ask a question.

ethanfrey:
btw, what project are you working on. do you have a team? some tight timelines, etc?

you seem to be looking a lot into tendermint-rs and ibc23 crates, and if you want things to move, it might be good to set up a call and see if someone could project manage this.

DaviRain-Su:
> btw, what project are you working on. do you have a team? some tight timelines, etc?
> 
> you seem to be looking a lot into tendermint-rs and ibc23 crates, and if you want things to move, it might be good to set up a call and see if someone could project manage this.

Oh, I'm just assisting in the processing of this library and I think informalsystem will upgrade the prost version later to solve this problem.

Issue #46, opened 2021-07-28
the trait `prost::Message` is not implemented for `ics23::ProofSpec`


DaviRain-Su:
> error[E0277]: the trait bound `ics23::ProofSpec: prost::Message` is not satisfied
  --> modules/src/ics23_commitment/specs.rs:40:13
   |
40 |             prost::Message::encode(ds, &mut encoded).unwrap();
   |             ^^^^^^^^^^^^^^^^^^^^^^ the trait `prost::Message` is not implemented for `ics23::ProofSpec`
   |
   = note: required by `prost::Message::encode`
error: aborting due to previous error
For more information about this error, try `rustc --explain E0277`.
error: could not compile `ibc`

When I open no-std I get an error.

This config: ics23 = { version = "0.6.5", default-features = false, git = "https://github.com/confio/ics23.git", branch = "master" }

project: https://github.com/octopus-network/ibc-rs/tree/dv-no-std

commit: https://github.com/octopus-network/ibc-rs/commit/40b4f6f48ad603556b9e9d0f3cf9ba1a4d5a4ea3

command: cargo check -p ibc

PR #48, opened 2021-07-30
Solutions : expected struct `TryFromIntError`, found struct `anyhow::Error`

DaviRain-Su:
> error[E0271]: type mismatch resolving `<u64 as TryFrom<usize>>::Error == anyhow::Error`
    --> /Users/davirain/.cargo/git/checkouts/ics23-fbd6714e79a75cb1/5a0206f/rust/src/ops.rs:71:28
     |
  71 |     let size: u64 = length.try_into()?;
     |                            ^^^^^^^^ expected struct `TryFromIntError`, found struct `anyhow::Error`

  >Compiling getrandom v0.2.3
  >For more information about this error, try `rustc --explain E0271`.
  error: could not compile `ics23` due to previous error
  warning: build failed, waiting for other jobs to finish...
  error: build failed

When I added ics23 to substrate for testing to see if the error reported by no_std was well supported.

### DIFF
--- a/rust/src/ops.rs
+++ b/rust/src/ops.rs
@@ -4,6 +4,7 @@ use ripemd160::Ripemd160;
 use sha2::{Digest, Sha256, Sha512, Sha512Trunc256};
 use sha3::Sha3_512;
 use std::convert::TryInto;
+use std::num::TryFromIntError;
 
 use crate::helpers::{Hash, Result};
 use crate::ics23::{HashOp, InnerOp, LeafOp, LengthOp};
@@ -68,7 +69,7 @@ fn do_length(length: LengthOp, data: &[u8]) -> Result<Hash> {
 }
 
 fn proto_len(length: usize) -> Result<Hash> {
-    let size: u64 = length.try_into()?;
+    let size: u64 = length.try_into().map_err(|e : TryFromIntError | anyhow::anyhow!(e))?;
     let mut len = Hash::new();
     prost::encoding::encode_varint(size, &mut len);
     Ok(len)


hu55a1n1:
Thanks for the PR @DaviRain-Su and thanks for this awesome library @ethanfrey! 🙏 
It would be great if we could merge this because the `ics23` crate fails to build in `no_std` mode currently and this PR fixes it. We're using this in our [ibc-rs](https://github.com/informalsystems/ibc-rs) project and this issue is blocking our release. 

Here's some context on the underlying cause of this issue -> 
> Since the ?-based error conversions would normally rely on the std::error::Error trait which is only available through std, no_std mode will require an explicit .map_err(Error::msg) when working with a non-Anyhow error type inside a function that returns Anyhow’s error type.

https://docs.rs/anyhow/1.0.44/anyhow/#no-std-support

ethanfrey:
@hu55a1n1 Thanks for bumping this issue and explaining why it was needed. It seems like follow up from https://github.com/confio/ics23/pull/41

Note the pr review and my request for proper testing and cleanup of the no_std work in  https://github.com/confio/ics23/issues/42

If you want to take over getting no_std work merged, I would be happy. I would ask.

1. rebase this pr. 
2. write at least one test to show the old code breaks, the new code works. If it is urgent, I can merge without tests, but prefer not to.
3. Commit to taking over #42 and possibly https://github.com/confio/ics23/issues/46 so the no_std work is properly tested and doesn't turn into urgent requests under someone else's release timeline.

I know little about no_std but happy to merge if things are properly tested and covered by ci.

ethanfrey:
Replaced by #54 (same code rebased)

PR #55, opened 2021-11-10
solidity implementation for ics23

lightyear15:
Code is splitted in 4 different libraries, following the layout of
functions and files of the go implementation
- Ops -> go/ops.go
- Proof -> go/proof.go
- Compress -> go/compress.go
- Ics23 -> go/ics23.go

Missing parts of the current implementation:
- sha512 hash option: solidity does not have built-in function for this
hash. A solidity implementation would utterly inefficient and very
costly (gas-wise)
- compress functions: the implementation will very likely be used to
verification, not for proof generation
- ProofSpec for tendermint and iavl definition: again, this
implementation is meant to be used to verify, not to provide proofs.

The implementation also provides a framework to run the same unit-test
cases that the go implementation has.
go-ethereum package provides a ``SimulatedBackend`` that can be used to
deploy a contract and call functions. UnitTest-specific contracts are
included in the libraries sol files.
The current go-ethereum version is the last with a working ``abigen``
command.

This commit also refactors the go project to make room for two go
modules in the same repository


lightyear15:
A few points to answer yet-to-ask reviewers' questions:
- This PR refactors the go implementation, there is no functional change, only refactoring tests code to being able to share the cases with solidity unit tests - also root-level go.mod has been removed
- I failed at automating the ``protoc`` for solidity build step. Unfortunately there is not yet official protobuf-solidity compiler/plugin. I used an [updated version](https://github.com/mkaczanowski/solidity-protobuf) of the [@datachainlab ](https://github.com/datachainlab/solidity-protobuf) compiler.
- even though ``go-ethereum`` pkg is now at 1.10.12. The latest working version of ``abigen`` is v.1.10.0
- circleCI does not provide any docker image with solidity tools like ``solc`` compiler. Using tools like truffle would have prevented me from reusing go unit test cases.

If this PR is too bulky and cumbersome to review, please let me know and I will try to split it into smaller, less scary pieces.

lightyear15:
@adizere, @andynog,
you might be interested in this.

ethanfrey:
I have yet to review, but 2 points.

1. Can you do the go stuff in another pr, separate from Solidity?

2. What is the license on the go-ethereum code you import? Much is GPL which is not good for this library.

lightyear15:
Hi @ethanfrey,
I understand that license mismatch is a problem. However, if we can't include any of the go-ethereum code, it would not make sense anymore to refactor the go code, as the only purpose of the changes was to import the test cases in the solidity unit test framework.
I can try to move to truffle framework which is licensed under MIT.

ethanfrey:
I have asked about licensing many times with the ICF and they are very admimant about Apache-2 only, no GPL family.

Since this package is imported by the Cosmos SDK, and thus affected every Cosmos blockchain, we need to be extra careful to maintain Apache 2 compliance.

ethanfrey:
Thank you for this work, and I look forward to seeing one based on MIT code

lightyear15:
Thanks @ethanfrey ,
I realised, I can move all the test cases description to json, so to be shared between go code and truffle's javascript code.
:v:

PR #57, opened 2021-11-23
Add `SmtSpec` for SMT proofs

roysc:
This will be needed for implementing [ADR-040](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-040-storage-and-smt-state-commitments.md) in Cosmos-SDK; specifically this PR depends on it: https://github.com/cosmos/cosmos-sdk/pull/10015


roysc:
This will need a Go version bump, should it go to 0.6.8 or would 0.7.0 be more semver-appropriate?

go/proof.go
```diff
@@ -40,6 +40,25 @@ var TendermintSpec = &ProofSpec{
 	},
 }
 
+// SmtSpec constrains the format for SMT proofs (as implemented by github.com/celestiaorg/smt)
```
ethanfrey, line 43: It would be nice to have a test case (even json file, table test... something generated by a real SMT that we can test against).

Let's verify this works with your SMT proofs.

ethanfrey (commented):
Thank you for adding this.

This is non-breaking addition, and actually could be copied directly in the app (no need to block on it). It is great to keep the standard specs used by most cosmos chains in this repo for easy reference. 

I would just ask for a test case to show this test properly verifying (and rejecting) at leats one good and one bad proof from SMT

go/proof.go
roysc: Originally this was in the cosmos-sdk repo, but I moved it here to keep a single source of truth.

Test cases are [here](https://github.com/cosmos/cosmos-sdk/pull/10015/files#diff-751f1a02f13499f5243d1b120677047e55f0ed907934676b1de420fab81dbd77), in the code for the sdk PR. I don't have one with a failed verification yet, but I will add that. Should I add any tests within this repo?

ethanfrey: Please add some basic tests in this repo.
Happy to have it here, just tested here as well

ethanfrey:
@roysc bumping

I am happy to merge as soon as there are a couple test specs.
There is a new JSON format for them an all now (thanks to the work on solidity), which might be easier to use cross-language. But even adding a positive test in Go would be nice.

roysc:
Sorry, I haven't had time to get to creating a proper serialization tool like `ics23-iavl` for the test data. I could look at just copying over some data as Go code in the meantime, but it's not particularly urgent yet to merge the spec into this repo, so I figured this could wait for the proper solution.

roysc:
After testing more thoroughly, I found that the proof verification algorithm needs to be updated to check for empty child nodes in non-existence proof paths to support SMT proofs - see PR at https://github.com/confio/ics23/pull/61.

I've also written a [ics23-smt ](https://github.com/roysc/ics23-smt) library to generate and check proof data (that depends on the above PR). I will rebase this branch on that fix and add the tests and data to complete it.

ethanfrey:
> After testing more thoroughly, I found that the proof verification algorithm needs to be updated to check for empty child nodes in non-existence proof paths to support SMT proofs

😄 See I'm not just being annoying when asking for test cases... they do find issues.

I will review #61 now

roysc:
Haha yes, the tests in the SDK were a bit too trivial it seems

ethanfrey:
Bumping this one to check the status.

Are you waiting for my review or am I waiting for more tests?

roysc:
This is also ready for review, sorry for the delay. The JS part is not complete, but I will leave it in for now.

roysc:
I've rebased and updated this now that the placeholder check is merged. I've also had to add a fix to the placeholder check logic, since it was being called with the wrong branch index. The fix is to detect the index from the padding within the empty check.

go/proof.go
```diff
@@ -230,7 +250,7 @@ func IsRightMost(spec *InnerSpec, path []*InnerOp) bool {
 
 	// ensure every step has a prefix and suffix defined to be rightmost, unless it is a placeholder node
 	for _, step := range path {
-		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !rightBranchesAreEmpty(spec, step, int32(last)) {
```
ethanfrey, line 233: good to remove

go/proof.go
```diff
@@ -310,12 +330,15 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
 }
 
-// leftBranchesAreEmpty returns true if the padding bytes correspond to all empty children
-// on the left side of this branch, ie. it's a valid placeholder on a leftmost path
-func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
-	idx := getPosition(spec.ChildOrder, branch)
+// leftBranchesAreEmpty returns true if the padding bytes correspond to all empty siblings
+// on the left side of a branch, ie. it's a valid placeholder on a leftmost path
+func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp) bool {
+	idx, err := orderFromPadding(spec, op)
```
ethanfrey, line 336: good improvement

rust/src/verify.rs
```diff
@@ -264,62 +264,48 @@ fn get_padding(spec: &ics23::InnerSpec, branch: i32) -> Result<Padding> {
...
-fn left_branches_are_empty(
-    spec: &ics23::InnerSpec,
-    op: &ics23::InnerOp,
-    branch: i32,
-) -> Result<bool> {
-    if let Some(&idx) = spec.child_order.iter().find(|&&x| x == branch) {
-        // count branches to left of this
-        let left_branches = idx as usize;
-        if left_branches == 0 {
+fn left_branches_are_empty(spec: &ics23::InnerSpec, op: &ics23::InnerOp) -> Result<bool> {
```
ethanfrey, line 267: looks cleaner

ethanfrey (approved):
looks good

PR #58, opened 2021-11-29
solidity implementation for ics23

lightyear15:
Code is splitted in 4 different libraries, following the layout of
functions and files of the go implementation
- Ops -> go/ops.go
- Proof -> go/proof.go
- Compress -> go/compress.go
- Ics23 -> go/ics23.go

Missing parts of the current implementation:
- sha512 hash option: solidity does not have built-in function for this
hash. A solidity implementation would be utterly inefficient and very
costly (gas-wise)
- compress functions: this implementation is meant be used for
verification, not for proof generation
- ProofSpec for tendermint and iavl definition: again, this
implementation is just to verify, not to generate
- Decompress functions are present but not working: currently
investigating a potention buf in the protobuf runtime
- Batch verification functions are temporarily disabled due to above


lightyear15:
Hello @adizee @andynog
you might be interested in this

@ethanfrey ,
Here is my PR with the solidity implementation.
I know I promised to use truffle, but after a few hours trying to write the first JS unitest I had a breakdown :smile:  , so I decided to move to [brownie](https://eth-brownie.readthedocs.io/en/stable/toctree.html) for a much better user experience. 
The framework it's released under MIT as well so I hope this change won't cause any license problem.

I am investigating the issue with the Protobuf decompress functions, trying to tailor the the simplest possible example for the maintainers of the [solidity-protobuf](https://github.com/datachainlab/solidity-protobuf) to start the investigation.

ethanfrey:
> Missing parts of the current implementation:

> sha512 hash option: solidity does not have built-in function for this
hash. A solidity implementation would be utterly inefficient and very
costly (gas-wise)

Agree. No know implementations use this. It was just added for completeness.

> compress functions: this implementation is meant be used for
verification, not for proof generation

Agree.

> ProofSpec for tendermint and iavl definition: again, this
implementation is just to verify, not to generate

The ProofSpec is needed to verify. You don't need to export it, but the test cases should have access to those proof specs so they can verify the same data. (You don't need a solidity object for it if you don't want)

> Decompress functions are present but not working: currently
investigating a potention buf in the protobuf runtime

These only apply to batch proofs. Current ICS / Relayer usage seems to just include a series of existence proofs and no batching/compression, so simply throwing an error "not yet implemented" and leaving an issue to add them seems fine

> Batch verification functions are temporarily disabled due to above

ethanfrey:
> The framework it's released under MIT as well so I hope this change won't cause any license problem.

MIT is good.

I will try to review soon, but it is fine to merge with the described functionality if it works

adlerjohn:
I will take a look at this, but might be a couple days.

lightyear15:
FYI:
With the help of the maintainers, I found out why ``decompress`` function doesn't work.
https://github.com/datachainlab/solidity-protobuf/issues/23
The current solidity-protobuf compiler does not support packed repeated fields.
As explained in the issue there is a workaround: adding attribute ``[packed=false]`` to field [``path``](https://github.com/confio/ics23/blob/42d3932593640f11cc39ac768592489b652d5edb/proofs.proto#L227)

... :thinking: ...
I wonder if this bug is also affecting message [``InnerSpec``](https://github.com/confio/ics23/blob/42d3932593640f11cc39ac768592489b652d5edb/proofs.proto#L181)

sol/tests/test_ics23compress.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@pytest.fixture
+def proof():
+    return accounts[0].deploy(Proof_UnitTest)
+
+
+@pytest.mark.parametrize("fname", [os.path.join(IAVL_TEST_DIR, fname) for fname in cases_fname])
+@pytest.mark.skip(reason="potential bug in ProtobufRuntime")
+def test_iavl_decompress(compress, proof, fname):
```
ethanfrey, line 26: Amazed at how expressive these tests can be with those decorators. Nice stuff.

ethanfrey: Shame they are skipped. But compress is not essential for v1

sol/tests/test_ics23ops.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.mark.parametrize("name,case", helpers.loadJsonTestCases("../testdata/TestLeafOpData.json"))
+def test_leafOp(ops, name, case):
+    if "sha-512" in name:
+        pytest.skip("sha-512 not supported in solidity")
+    opCase = convert.leafOp(case["Op"])
+    key = base64.b64decode(case["Key"] or "")
+    value = base64.b64decode(case["Value"] or "")
+    if case["IsErr"]:
+        with brownie.reverts():
```
ethanfrey, line 23: Does this ensure it throws an exception?

sol/tests/test_ics23vectors.py
```diff
@@ -0,0 +1,47 @@
...
+def ics23():
+    return accounts[0].deploy(ICS23_UnitTest)
+
+
+@pytest.fixture
+def proof():
+    return accounts[0].deploy(Proof_UnitTest)
+
+
+@pytest.mark.parametrize("fname", [os.path.join(IAVL_TEST_DIR, fname) for fname in cases_fname])
```
ethanfrey, line 24: looks good

ethanfrey (approved):
The test code looks nice and proper.

I cannot review solidity, so I would take someone else on that, but from my end, we could merge this as is (tested) and then have another pass with a solidity reviewer to ensure there are no issues in that code.

ethanfrey:
@lightyear15 sorry for the long delay.

If you want, I can merge it. Just let me know.
Otherwise, feel free to get a reviewer for the solidity stuff first

sol/tests/test_ics23ops.py
lightyear15: [Yes](https://eth-brownie.readthedocs.io/en/stable/api-test.html?highlight=revertcontext#revertcontextmanager)

lightyear15:
Hello @ethanfrey ,
thanks a lot for your review.
I will talk to ICF and see how they want to proceed

sol/contracts/ics23Proof.sol
```diff
@@ -0,0 +1,229 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.2;
+pragma experimental ABIEncoderV2;
```
mkaczanowski, line 3: Pragma `ABIEncoderV2` is not required since 0.8 release, I think we can remove it

> You can choose to use the old behaviour using pragma abicoder v1;. The pragma pragma experimental ABIEncoderV2; is still valid, but it is deprecated and has no effect. If you want to be explicit, please use pragma abicoder v2; instead.

Source ^^: https://docs.soliditylang.org/en/v0.8.0/080-breaking-changes.html

sol/contracts/ics23Ops.sol
```diff
@@ -0,0 +1,166 @@
...
+
+library Ops {
+
+    // LeafOp operations
+    function applyOp(LeafOp.Data memory leafOp, bytes memory key, bytes memory value) internal pure returns(bytes memory) {
+        require(key.length > 0); // dev: Leaf op needs key
+        require(value.length > 0); // dev: Leaf op needs value
+        bytes memory pKey = prepareLeafData(leafOp.prehash_key, leafOp.length, key);
+        bytes memory pValue = prepareLeafData(leafOp.prehash_value, leafOp.length, value);
+        bytes memory data = bytes.concat(leafOp.prefix, pKey, pValue);
```
mkaczanowski, line 18: `bytes.concat` [comes with 0.8.4](https://docs.soliditylang.org/en/v0.8.4/types.html#bytes-concat-function). This code doesn't compile if you use 0.8.2 compiler.

I'd replace it with `abi.encodePacked` for now.

sol/contracts/ics23.sol
```diff
@@ -0,0 +1,107 @@
...
+import {BatchProof, CompressedBatchProof, CommitmentProof, ProofSpec, ExistenceProof, NonExistenceProof} from "./proofs.sol";
+import {Compress} from "./ics23Compress.sol";
+import {Proof} from "./ics23Proof.sol";
+import {Ops} from "./ics23Ops.sol";
+import {BytesLib} from "GNSPS/solidity-bytes-utils@0.8.0/contracts/BytesLib.sol";
+
+library Ics23  {
+
+    // verifyMembership, throws an exception in case anything goes wrong
+    function verifyMembership(ProofSpec.Data memory spec, bytes memory commitmentRoot, CommitmentProof.Data memory proof, bytes memory key, bytes memory value) internal pure {
```
mkaczanowski, line 12: the function should return bool. If the proof is invalid, that's okay, let the user handle it. `require` will bail out and revert the execution

lightyear15: I see I see :thinking: 
yeah it's a legit request 
I can refactor all the functions to return a bool as second element in sort of a Go error handling style...
but then we would lose the information about the error :thinking: 
Maybe I can write an enum to represent the error, there would be a lot of enums, but at least we wouldn't lose information about the error...

sol/contracts/ics23.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    function isRight(ExistenceProof.Data memory right, bytes memory key) private pure returns(bool) {
+        // ExistenceProof.isNil does not work
+        return ExistenceProof._empty(right) || Ops.compare(right.key, key) > 0;
+    }
+}
+
+
+contract ICS23_UnitTest {
```
mkaczanowski, line 91: I don't know what is considered to be the best practice in solidity, but by looking at the [openzeppelin repo](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/mocks) it seems that keeping the tests under the `mocks` repository is the thing.

I kinda like ^^ that, because the `contracts` and `libraries` are explicitly separated, wdyt?

sol/contracts/ics23.sol
lightyear15: Hi @mkaczanowski ,
my last commit changes the return values of functions.
Function callers will get an enum encoding the Error, no more require and revert except those in the auto-generated bindings

sol/contracts/ProtoBufRuntime.sol
```diff
@@ -0,0 +1,2990 @@
+// SPDX-License-Identifier: Apache-2.0
```
adlerjohn, line 1: The biggest issue here is the use of this library. It is a protobuf codec that may or may not follow a non-existent protobuf spec, not an [ADR-027](https://github.com/cosmos/cosmos-sdk/blob/b0d3ef96506054b66bc1160d48f4b24105e7d35a/docs/architecture/adr-027-deterministic-protobuf-serialization.md) codec. The latter is highly recommended (read: required) to avoid the possibility of malleability attacks such as [this one](https://medium.com/immunefi/polygon-double-spend-bug-fix-postmortem-2m-bounty-5a1db09db7f1). You should be using the following libraries to deal with protobuf.

1. https://github.com/celestiaorg/protobuf3-solidity
2. https://github.com/celestiaorg/protobuf3-solidity-lib

sol/contracts/ics23Proof.sol
```diff
@@ -0,0 +1,307 @@
...
+
+    enum VerifyExistenceError{
+        None,
+        KeyNotMatching,
+        ValueNotMatching,
+        CheckSpec,
+        CalculateRoot,
+        RootNotMatching
+    }
+    // ExistenceProof
```
adlerjohn, line 22: These library internal methods should all be documented with [NatSpec](https://docs.soliditylang.org/en/v0.8.2/natspec-format.html).

adlerjohn (changes requested):

sol/contracts/ProtoBufRuntime.sol
lightyear15: HI @adlerjohn ,
the problem with the ``celestiaorg`` compiler is that, as reported in the README, it does not support ``oneof``  needed in [``BatchEntry``](https://github.com/confio/ics23/blob/e736db6536ced19ed677b4d4d00ade63d47d4a3d/proofs.proto#L199)

adlerjohn: `oneof` can be implemented as a library! Much better than fixing the currently-used library to be ADR-027-compliant. PRs welcome.

lightyear15: I have talked to the [datachainlab/solidity-protobuf](https://github.com/datachainlab/solidity-protobuf) compiler's maintainers.
They have plans to make it ADR-27 compliant soon.
I think I will just sit and wait for it :grin:

lightyear15:
Datachain lab have updated their compiler which now is ADR-27 compliant.
The changes to the branch updated the compiled sol libraries with the new version of the compiler.
This PR should be ready for review as all the @adlerjohn comments have been addressed

schnetzlerjoe:
@ethanfrey is work on this complete? Curious!

hu55a1n1:
Hi @adlerjohn and @mkaczanowski, would you be able to take a look to see if your comments have been addressed and approve the PR if you find the time?

crodriguezvega:
Is there still interest in merging this PR?

lightyear15:
> Is there still interest in merging this PR?
 
 not on my side

crodriguezvega:
> not on my side

Ok, thank you. Then I will close the PR for now.

PR #59, opened 2021-12-02
Add buf.build

tac0turtle:
But provides a layout to the repo that is common amongst Protobuf repos. This helps with registering proto types in the proto registry. 

In ibc, proofs.proto is imported via confio/proofs.proto, which fails in grace reflection as its not correct in the proto registry 

closes https://github.com/confio/ics23/issues/32


proto/confio/v1/proofs.proto
```diff
@@ -2,15 +2,17 @@ syntax = "proto3";
 
 package ics23;
```
marbar3778, line 3: I would propose changing this to condo.v1.ics23 to have version proto files. thoughts?

proto/confio/v1/proofs.proto
```diff
@@ -1,16 +1,18 @@
 syntax = "proto3";
 
-package ics23;
+package confio.v1.ics23;
```
marbar3778, line 3: this will need to be regenerated, want to double check if its okay before doing so. 

DO NOT MERGE

ethanfrey: Sure, we can rename this.
Let's make sure this gets into a 0.7.0 release as it is breaking for the client.
I assume the wire format doesn't change, which we do try to avoid.

ethanfrey (commented):
Looks like a nice cleanup.
Curious why the CI doesn't run

proto/confio/v1/proofs.proto
marbar3778: correct wire format does not change. Would you be up for cutting 1.0?

marbar3778:
this is ready for go. For js, its unclear to me what I need to run? Rust, is panicking

ethanfrey:
> this is ready for go. For js, its unclear to me what I need to run? Rust, is panicking

Ooh, sounds like fun. I will come back to this when I have a bit of time.

PR #61, opened 2021-12-12
Check for placeholder nodes in proof verification

roysc:
Certain proofs (like those for a SMT) can contain empty nodes which are equal to the `EmptyChild`, but currently these will cause verification to fail for non-existence proofs if they appear where a real sibling node would be invalid, i.e. on a path that must be the leftmost or rightmost path of a subtree.

This adds a check to detect and ignore such nodes during verification of left- and rightmost paths. This will be needed in order to support SMT proofs in https://github.com/confio/ics23/pull/57.

### DIFF
--- a/go/proof.go
+++ b/go/proof.go
@@ -314,7 +314,7 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
 func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 	idx := getPosition(spec.ChildOrder, branch)
 	// compare the prefix bytes with the appropriate number of empty children
-	leftChildren := len(spec.ChildOrder) - 1 - idx
+	leftChildren := idx
 	actualPrefix := len(op.Prefix) - leftChildren*int(spec.ChildSize)
 	if actualPrefix < 0 {
 		return false
@@ -332,11 +332,12 @@ func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 // on the right side of this branch, ie. it's a valid placeholder on a rightmost path
 func rightBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 	idx := getPosition(spec.ChildOrder, branch)
+	rightBranches := len(spec.ChildOrder) - 1 - idx
 	// compare the suffix bytes with the appropriate number of empty children
-	if len(op.Suffix) != idx*int(spec.ChildSize) {
+	if len(op.Suffix) != rightBranches*int(spec.ChildSize) {
 		return false
 	}
-	for i := 0; i < idx; i++ {
+	for i := 0; i < rightBranches; i++ {
 		from := i * int(spec.ChildSize)
 		if !bytes.Equal(spec.EmptyChild, op.Suffix[from:from+int(spec.ChildSize)]) {
 			return false
--- a/go/proof.go
+++ b/go/proof.go
@@ -210,35 +210,27 @@ func (p *NonExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []b
 	return nil
 }
 
-// IsLeftMost returns true if this is the left-most path in the tree
+// IsLeftMost returns true if this is the left-most path in the tree, excluding placeholder (empty child) nodes
 func IsLeftMost(spec *InnerSpec, path []*InnerOp) bool {
 	minPrefix, maxPrefix, suffix := getPadding(spec, 0)
 
-	// ensure every step has a prefix and suffix defined to be leftmost
+	// ensure every step has a prefix and suffix defined to be leftmost, unless it is a placeholder node
 	for _, step := range path {
-		if !hasPadding(step, minPrefix, maxPrefix, suffix) {
-			// if this is a placeholder node, skip it
-			if leftBranchesAreEmpty(spec, step, 0) {
-				continue
-			}
+		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !leftBranchesAreEmpty(spec, step, 0) {
 			return false
 		}
 	}
 	return true
 }
 
-// IsRightMost returns true if this is the left-most path in the tree
+// IsRightMost returns true if this is the left-most path in the tree, excluding placeholder (empty child) nodes
 func IsRightMost(spec *InnerSpec, path []*InnerOp) bool {
 	last := len(spec.ChildOrder) - 1
 	minPrefix, maxPrefix, suffix := getPadding(spec, int32(last))
 
-	// ensure every step has a prefix and suffix defined to be rightmost
+	// ensure every step has a prefix and suffix defined to be rightmost, unless it is a placeholder node
 	for _, step := range path {
-		if !hasPadding(step, minPrefix, maxPrefix, suffix) {
-			// if this is a placeholder node, skip it
-			if rightBranchesAreEmpty(spec, step, int32(last)) {
-				continue
-			}
+		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !rightBranchesAreEmpty(spec, step, int32(last)) {
 			return false
 		}
 	}
--- a/go/proof.go
+++ b/go/proof.go
@@ -217,6 +217,10 @@ func IsLeftMost(spec *InnerSpec, path []*InnerOp) bool {
 	// ensure every step has a prefix and suffix defined to be leftmost
 	for _, step := range path {
 		if !hasPadding(step, minPrefix, maxPrefix, suffix) {
+			// if this is a placeholder node, skip it
+			if leftBranchesAreEmpty(spec, step, 0) {
+				continue
+			}
 			return false
 		}
 	}
@@ -231,6 +235,10 @@ func IsRightMost(spec *InnerSpec, path []*InnerOp) bool {
 	// ensure every step has a prefix and suffix defined to be rightmost
 	for _, step := range path {
 		if !hasPadding(step, minPrefix, maxPrefix, suffix) {
+			// if this is a placeholder node, skip it
+			if rightBranchesAreEmpty(spec, step, int32(last)) {
+				continue
+			}
 			return false
 		}
 	}
@@ -307,6 +315,42 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
 	// count how many children are in the suffix
 	suffix = (len(spec.ChildOrder) - 1 - idx) * int(spec.ChildSize)
 	return
+}
+
+// leftBranchesAreEmpty returns true if the padding bytes correspond to all empty children
+// on the left side of this branch, ie. it's a valid placeholder on a leftmost path
+func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
+	idx := getPosition(spec.ChildOrder, branch)
+	// compare the prefix bytes with the appropriate number of empty children
+	leftChildren := len(spec.ChildOrder) - 1 - idx
+	actualPrefix := len(op.Prefix) - leftChildren*int(spec.ChildSize)
+	if actualPrefix < 0 {
+		return false
+	}
+	for i := 0; i < leftChildren; i++ {
+		from := actualPrefix + i*int(spec.ChildSize)
+		if !bytes.Equal(spec.EmptyChild, op.Prefix[from:from+int(spec.ChildSize)]) {
+			return false
+		}
+	}
+	return true
+}
+
+// rightBranchesAreEmpty returns true if the padding bytes correspond to all empty children
+// on the right side of this branch, ie. it's a valid placeholder on a rightmost path
+func rightBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
+	idx := getPosition(spec.ChildOrder, branch)
+	// compare the suffix bytes with the appropriate number of empty children
+	if len(op.Suffix) != idx*int(spec.ChildSize) {
+		return false
+	}
+	for i := 0; i < idx; i++ {
+		from := i * int(spec.ChildSize)
+		if !bytes.Equal(spec.EmptyChild, op.Suffix[from:from+int(spec.ChildSize)]) {
+			return false
+		}
+	}
+	return true
 }
 
 // getPosition checks where the branch is in the order and returns
--- a/go/proof.go
+++ b/go/proof.go
@@ -281,10 +281,15 @@ func isLeftStep(spec *InnerSpec, left *InnerOp, right *InnerOp) bool {
 		panic(err)
 	}
 
+	// for idx := leftidx + 1; idx != rightidx; idx++ {
+	// leftBranchesAreEmpty()
+	// }
+
 	// TODO: is it possible there are empty (nil) children???
 	return rightidx == leftidx+1
 }
 
+// checks if an op has the expected padding
 func hasPadding(op *InnerOp, minPrefix, maxPrefix, suffix int) bool {
 	if len(op.Prefix) < minPrefix {
 		return false
@@ -313,13 +318,17 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
 // on the left side of this branch, ie. it's a valid placeholder on a leftmost path
 func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 	idx := getPosition(spec.ChildOrder, branch)
-	// compare the prefix bytes with the appropriate number of empty children
-	leftChildren := idx
-	actualPrefix := len(op.Prefix) - leftChildren*int(spec.ChildSize)
+	// count branches to left of this
+	leftBranches := idx
+	if leftBranches == 0 {
+		return false
+	}
+	// compare prefix with the expected number of empty branches
+	actualPrefix := len(op.Prefix) - leftBranches*int(spec.ChildSize)
 	if actualPrefix < 0 {
 		return false
 	}
-	for i := 0; i < leftChildren; i++ {
+	for i := 0; i < leftBranches; i++ {
 		from := actualPrefix + i*int(spec.ChildSize)
 		if !bytes.Equal(spec.EmptyChild, op.Prefix[from:from+int(spec.ChildSize)]) {
 			return false
@@ -332,10 +341,14 @@ func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 // on the right side of this branch, ie. it's a valid placeholder on a rightmost path
 func rightBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 	idx := getPosition(spec.ChildOrder, branch)
+	// count branches to right of this one
 	rightBranches := len(spec.ChildOrder) - 1 - idx
-	// compare the suffix bytes with the appropriate number of empty children
-	if len(op.Suffix) != rightBranches*int(spec.ChildSize) {
+	if rightBranches == 0 {
 		return false
+	}
+	// compare suffix with the expected number of empty branches
+	if len(op.Suffix) != rightBranches*int(spec.ChildSize) {
+		return false // sanity check
 	}
 	for i := 0; i < rightBranches; i++ {
 		from := i * int(spec.ChildSize)
--- a/go/proof.go
+++ b/go/proof.go
@@ -281,10 +281,6 @@ func isLeftStep(spec *InnerSpec, left *InnerOp, right *InnerOp) bool {
 		panic(err)
 	}
 
-	// for idx := leftidx + 1; idx != rightidx; idx++ {
-	// leftBranchesAreEmpty()
-	// }
-
 	// TODO: is it possible there are empty (nil) children???
 	return rightidx == leftidx+1
 }
--- a/go/proof_data_test.go
+++ b/go/proof_data_test.go
@@ -105,7 +105,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 			},
 			Spec:    SpecWithEmptyChild,
 			IsLeft:  true,
-			IsRight: false,
+			IsRight: true,
 		},
 		EmptyBranchTestStruct{
 			Op: &InnerOp{
@@ -114,7 +114,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 				Hash:   SpecWithEmptyChild.InnerSpec.Hash,
 			},
 			Spec:    SpecWithEmptyChild,
-			IsLeft:  false,
+			IsLeft:  true,
 			IsRight: true,
 		},
 		// non-empty cases
@@ -126,7 +126,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 			},
 			Spec:    SpecWithEmptyChild,
 			IsLeft:  false,
-			IsRight: false,
+			IsRight: true,
 		},
 		EmptyBranchTestStruct{
 			Op: &InnerOp{
@@ -135,7 +135,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 				Hash:   SpecWithEmptyChild.InnerSpec.Hash,
 			},
 			Spec:    SpecWithEmptyChild,
-			IsLeft:  false,
+			IsLeft:  true,
 			IsRight: false,
 		},
 		EmptyBranchTestStruct{
@@ -146,7 +146,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 			},
 			Spec:    SpecWithEmptyChild,
 			IsLeft:  false,
-			IsRight: false,
+			IsRight: true,
 		},
 		EmptyBranchTestStruct{
 			Op: &InnerOp{
@@ -155,7 +155,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 				Hash:   SpecWithEmptyChild.InnerSpec.Hash,
 			},
 			Spec:    SpecWithEmptyChild,
-			IsLeft:  false,
+			IsLeft:  true,
 			IsRight: false,
 		},
 		// some cases using a spec with no empty child
@@ -167,7 +167,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 			},
 			Spec:    TendermintSpec,
 			IsLeft:  false,
-			IsRight: false,
+			IsRight: true,
 		},
 		EmptyBranchTestStruct{
 			Op: &InnerOp{
@@ -176,7 +176,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 				Hash:   TendermintSpec.InnerSpec.Hash,
 			},
 			Spec:    TendermintSpec,
-			IsLeft:  false,
+			IsLeft:  true,
 			IsRight: false,
 		},
 	}
--- a/go/proof_data_test.go
+++ b/go/proof_data_test.go
@@ -105,7 +105,7 @@ func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
 			},
 			Spec:    SpecWithEmptyChild,
 			IsLeft:  true,
-			IsRight: true,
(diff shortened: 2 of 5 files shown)


go/proof.go
```diff
@@ -217,6 +217,10 @@ func IsLeftMost(spec *InnerSpec, path []*InnerOp) bool {
 	// ensure every step has a prefix and suffix defined to be leftmost
 	for _, step := range path {
 		if !hasPadding(step, minPrefix, maxPrefix, suffix) {
+			// if this is a placeholder node, skip it
+			if isPlaceholder(spec, 0, step, true) {
```
ethanfrey, line 221: These calls make sense.

Since most of the logic is inside an if statement, why not make two functions - `areLeftBranchesEmpty` and `areRightBranchesEmpty` or something like that?

go/proof.go
```diff
@@ -309,6 +317,40 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
+	idx := getPosition(spec.ChildOrder, branch)
+
+	// compare the prefix bytes with the appropriate number of empty children
+	if inPrefix {
+		prefixCt := len(spec.ChildOrder) - 1 - idx
+		lenDiff := len(op.Prefix) - prefixCt*int(spec.ChildSize)
+		if lenDiff < int(spec.MinPrefixLength) || lenDiff > int(spec.MaxPrefixLength) {
+			return false
+		}
+		for i := 0; i < prefixCt; i++ {
```
ethanfrey, line 332: I get what you are trying to do, but I am unsure if this is correct.

Definitely could use some test cases... first a few hand-made examples just combining bytes and unit testing this method. Then a higher level on (in #57) that applies it to your real case.

go/proof.go
```diff
@@ -309,6 +317,40 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
+			// walk back from the end of the prefix, since its length could vary
+			from := len(op.Prefix) - (prefixCt-i)*int(spec.ChildSize)
+			if bytes.Compare(spec.EmptyChild, op.Prefix[from:from+int(spec.ChildSize)]) != 0 {
```
ethanfrey, line 335: Very good use of https://github.com/confio/ics23/blob/53a3a58ab8d224eef792845a65c7e8fc66712b7d/proofs.proto#L185-L186

I realise now I never check this anywhere, and this was a vestige of a spec indended to cover that case, but never implemented.

ethanfrey (commented):
Thanks for this contribution. Definitely a reason to make a 0.7 bump.

I would ask for a little cleanup and some unit tests on this, then I can approve.

I would love to see an implementation in js and/or Rust if you can... otherwise, I guess I need to make them before merge to keep them all in sync.

That said, once tests look good and I approve, please make #57 on top of this as the Go code will not change anymore.

go/proof.go
roysc: when I started writing the function, I thought the paths might share some code - but yes, I can just split it up.

go/proof.go
roysc: That's what I figured. Glad it was ready to use for this!

roysc:
Will add tests today, and port to the other languages. #57 is rebased on this, so this code is at least tested on those generated cases, but of course I'll add unit tests too, and maybe some proof tests.

ethanfrey:
> Will add tests today, and port to the other languages. #57 is rebased on this, so this code is at least tested on those generated cases, but of course I'll add unit tests too, and maybe some proof tests.

Thank you

roysc:
I've added unit tests and ported the code to Rust. I took a stab at adding the test data as JSON and deserializing in both languages, but I'm not sure how to cleanly match Go's base64 serialization in Rust, so for now, I've just hard coded and ported the test cases.

I also worked on porting the code to Typescript, but I'm not familiar with Node. I may need some help or extra time on that, or maybe it can be added in a separate PR.

go/proof.go
```diff
@@ -217,6 +217,10 @@ func IsLeftMost(spec *InnerSpec, path []*InnerOp) bool {
 	// ensure every step has a prefix and suffix defined to be leftmost
```
robert-zaremba, line 217: we need to update the function comment, and be more specific.

go/proof.go
```diff
@@ -231,6 +235,10 @@ func IsRightMost(spec *InnerSpec, path []*InnerOp) bool {
 	// ensure every step has a prefix and suffix defined to be rightmost
```
robert-zaremba, line 235: need to update function doc string as well

robert-zaremba (commented):
utACK. Need to update doc strings.

robert-zaremba:
We should update the JS / TS implementation as well.

ethanfrey:
> I've added unit tests and ported the code to Rust. I took a stab at adding the test data as JSON and deserializing in both languages, but I'm not sure how to cleanly match Go's base64 serialization in Rust, so for now, I've just hard coded and ported the test cases.
> 
> I also worked on porting the code to Typescript, but I'm not familiar with Node. I may need some help or extra time on that, or maybe it can be added in a separate PR.

Thank you for taking that on. I will take a quick stab at this.

go/proof.go
```diff
@@ -309,6 +317,42 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
 	return
 }
 
+// leftBranchesAreEmpty returns true if the padding bytes correspond to all empty children
```
ethanfrey, line 320: Both functions look quite clear and correct (name and implementation).

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+func EmptyBranchTestData(t *testing.T) []EmptyBranchTestStruct {
+	return []EmptyBranchTestStruct{
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: []byte{1},
+				Suffix: InnerSpecWithEmptyChild.EmptyChild,
+			},
+			Spec:   &InnerSpecWithEmptyChild,
+			IsTrue: true,
+			IsLeft: false,
```
ethanfrey, line 98: This is both left and right... I would like to see a copy of this with IsLeft:true as well

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+	IsTrue bool
+	IsLeft bool
+}
+
+var InnerSpecWithEmptyChild = InnerSpec{
+	ChildOrder:      []int32{0, 1},
+	ChildSize:       32,
+	MinPrefixLength: 1,
+	MaxPrefixLength: 1,
+	EmptyChild:      []byte("32_empty_child_placeholder_bytes"),
```
ethanfrey, line 85: nice to use something besides 0s for tests

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+				Prefix: []byte{1},
+				Suffix: InnerSpecWithEmptyChild.EmptyChild,
+			},
+			Spec:   &InnerSpecWithEmptyChild,
+			IsTrue: true,
+			IsLeft: false,
+		},
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: append([]byte{1}, make([]byte, 32)...),
```
ethanfrey, line 102: I'm confused... this should be invalid.

The prefix is 33 bytes (so one child on the left) and the suffix is 32 bytes (one child on the right). This is too long

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+				Suffix: InnerSpecWithEmptyChild.EmptyChild,
+			},
+			Spec:   &InnerSpecWithEmptyChild,
+			IsTrue: true,
+			IsLeft: false,
+		},
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: []byte{1},
+				Suffix: make([]byte, 32),
```
ethanfrey, line 112: nice assert this is not considered empty

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+			},
+			Spec:   &InnerSpecWithEmptyChild,
+			IsTrue: false,
+			IsLeft: false,
+		},
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: []byte{1},
```
ethanfrey, line 120: again, this is invalid.

maybe add another check if it even matches the spec, but here I would assume we only want to focus on valid ops and just test the left/right value.

Please verify the spec with `op.CheckAgainstSpec(InnerSpecWithEmptyChild)`. And assert that is true for all these cases before the other checks. 

Maybe you can pull out the invalid tests case to another test that just checks validity (and add a valid case or two there?)

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+				Suffix: append(InnerSpecWithEmptyChild.EmptyChild, []byte("xxxx")...),
+			},
+			Spec:   &InnerSpecWithEmptyChild,
+			IsTrue: false,
+			IsLeft: false,
+		},
+		EmptyBranchTestStruct{
```
ethanfrey, line 136: Likewise invalid Op, doesn't meet spec.

go/proof_data_test.go
```diff
@@ -69,3 +69,135 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+			Op: &InnerOp{
+				Prefix: []byte{1},
+				Suffix: append(InnerSpecWithEmptyChild.EmptyChild, []byte("xxxx")...),
+			},
+			Spec:   &InnerSpecWithEmptyChild,
+			IsTrue: false,
```
ethanfrey, line 133: I would actually do `IsRight` and `IsLeft` and check them both.
In some cases it is right, others left, others both left and right. Just changing the meaning here let's us assert all that cleaner

go/proof_test.go
```diff
@@ -54,3 +54,21 @@ func TestCheckAgainstSpec(t *testing.T) {
 		})
 	}
 }
+
+func TestEmptyBranch(t *testing.T) {
+	cases := EmptyBranchTestData(t)
+
+	for i, tc := range cases {
+		var res bool
+		if tc.IsLeft {
```
ethanfrey, line 63: eg.

```go
if leftBranchesAreEmpty(tc.Spec, tc.Op, 0) != spec.IsLeft {
  t.Errorf("Expected leftBranchesAreEmpty to be %b but it wasn't", spec.IsLeft)
}
if rightBranchesAreEmpty(tc.Spec, tc.Op, 0) != spec.IsRight {
  t.Errorf("Expected rightBranchesAreEmpty to be %b but it wasn't", spec.IsRight)
}
```

rust/src/verify.rs
```diff
@@ -170,6 +170,9 @@ fn ensure_left_most(spec: &ics23::InnerSpec, path: &[ics23::InnerOp]) -> Result<
     let pad = get_padding(spec, 0)?;
     for step in path {
         if !has_padding(step, &pad) {
+            if left_branches_are_empty(spec, step, 0)? {
```
ethanfrey, line 173: Maybe the same applies to Go, but looking at this I think it could be simplified:

``rust
if !has_padding(step, &pad) && !left_branches_are_empty(spec, step, 0)? {
    bail!("step not leftmost");
}
```

File: rust/src/verify.rs [hunk 16]
```diff
@@ -258,11 +264,61 @@ fn get_padding(spec: &ics23::InnerSpec, branch: i32) -> Result<Padding> {
...
+    op: &ics23::InnerOp,
+    branch: i32,
+) -> Result<bool> {
+    if let Some(&idx) = spec.child_order.iter().find(|&&x| x == branch) {
+        // compare the prefix bytes with the appropriate number of empty children
+        let left_children = spec.child_order.len() - 1 - idx as usize;
+        if op.prefix.len() < left_children * spec.child_size as usize {
+            return Ok(false);
+        }
+        let actual_prefix = op.prefix.len() - left_children * spec.child_size as usize;
```
ethanfrey, line 280: I think you could use `.checked_sub().ok_or(SomeError)` and avoid the if statement above (handle it here), but this is fine

rust/src/verify.rs
```diff
@@ -526,4 +582,186 @@ mod tests {
...
+                    hash: non_empty_spec.hash,
+                },
+                spec: &non_empty_spec,
+                is_true: false,
+                is_left: true,
+            },
+        ];
+
+        for (i, case) in cases.iter().enumerate() {
+            let res = if case.is_left {
```
ethanfrey, line 754: Same comments as for the go code. Both to cover both left and right at once as well as to avoid checking invalid ops.

rust/src/verify.rs
```diff
@@ -526,4 +582,186 @@ mod tests {
...
+                    suffix: vec![],
+                    hash: spec.hash,
+                },
+                spec: spec,
+                is_true: true,
+                is_left: true,
+            },
+            EmptyBranchCase {
+                op: ics23::InnerOp {
+                    prefix: vec![1_u8]
```
ethanfrey, line 693: Works, but I think `[&[1u8], &spec.empty_child].concat().to_vec()` will do the same and is a bit easier to read.

ethanfrey (commented):
Thank you for this.

Besides the comments by Robert, I added some more, mainly on the test code.

If you can make those adjustments, I am happy to merge it and I will take care of the TypeScript implementation.

ethanfrey:
I was going to add a commit, then realized it was on a fork, and better just to merge this, then I can make a new branch

roysc:
Thanks, I will push what I have for Typescript so far, it should be usable. I just didn't want to delay this while I study up on the node build system and testing.

Made a draft PR - https://github.com/confio/ics23/pull/62

go/proof_data_test.go
roysc: That's true, but these functions aren't checking the total padding, as that's already done before it's called - they just check one side. I can remove these cases since they don't represent real usage.

go/proof_data_test.go
roysc: I tried making a new test for inner op validity, but currently `(InnerOp) CheckAgainstSpec` doesn't check for total byte length. So I added that check, but that causes `TestCheckAgainstSpec` to fail all the "valid" cases, which actually use[ inner ops](https://github.com/confio/ics23/blob/e736db6536ced19ed677b4d4d00ade63d47d4a3d/testdata/TestCheckAgainstSpecData.json#L13) with invalid prefix bytes. I could change those, but it seems out of scope for this PR. I will just remove the invalid cases here.

ethanfrey:
Okay, this is blocking #57 right?

I am mostly back from holidays, I will take a look this week or early next week. Please ping me here

roysc:
Alright, I think the Go and Rust work on this is complete. @ethanfrey could you review again, and if it looks good let's move on to the TS port.

go/proof_data_test.go
```diff
@@ -69,3 +69,115 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+			IsRight: false,
+		},
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: []byte{1},
+				Suffix: emptyChild,
+				Hash:   SpecWithEmptyChild.InnerSpec.Hash,
+			},
+			Spec:    SpecWithEmptyChild,
+			IsLeft:  false,
```
ethanfrey, line 117: Likewise, this is leftmost from the previous definition

go/proof_data_test.go
```diff
@@ -69,3 +69,115 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+		// non-empty cases
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: append([]byte{1}, make([]byte, 32)...),
+				Suffix: nil,
+				Hash:   SpecWithEmptyChild.InnerSpec.Hash,
+			},
+			Spec:    SpecWithEmptyChild,
+			IsLeft:  false,
+			IsRight: false,
```
ethanfrey, line 129: Expect: IsRight: true

go/proof_data_test.go
```diff
@@ -69,3 +69,115 @@ func CheckAgainstSpecTestData(t *testing.T) map[string]CheckAgainstSpecTestStruc
...
+		},
+		EmptyBranchTestStruct{
+			Op: &InnerOp{
+				Prefix: []byte{1},
+				Suffix: make([]byte, 32),
+				Hash:   SpecWithEmptyChild.InnerSpec.Hash,
+			},
+			Spec:    SpecWithEmptyChild,
+			IsLeft:  false,
```
ethanfrey, line 138: Expect: IsLeft: true

go/proof_test.go
```diff
@@ -54,3 +54,21 @@ func TestCheckAgainstSpec(t *testing.T) {
...
+
+func TestEmptyBranch(t *testing.T) {
+	cases := EmptyBranchTestData(t)
+
+	for _, tc := range cases {
+		t.Run("case", func(t *testing.T) {
+			if err := tc.Op.CheckAgainstSpec(tc.Spec); err != nil {
+				t.Errorf("Invalid InnerOp: %v", err)
+			}
+			if leftBranchesAreEmpty(tc.Spec.InnerSpec, tc.Op, 0) != tc.IsLeft {
```
ethanfrey, line 66: Hmm, you are calling `leftBranchesAreEmpty` not `IsLeftMost`, so maybe my comments are wrong above.

However if there are no left branch, I would assume that `leftBranchesAreEmpty` would be true

go/proof.go
```diff
@@ -309,6 +309,42 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
+// on the left side of this branch, ie. it's a valid placeholder on a leftmost path
+func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
+	idx := getPosition(spec.ChildOrder, branch)
+	// compare the prefix bytes with the appropriate number of empty children
+	leftChildren := len(spec.ChildOrder) - 1 - idx
```
ethanfrey, line 317: Looking at the failing test cases, I think this math is wrong.

Assume there are 2 children, those would have position 0, and 1.

For position 0, I would get 2-1-0 = 1 left children (but should be 0)
For position 1, I would get 2-1-1 = 0 left children. (but should be 1)

I think `leftChildren := idx` would be correct

go/proof.go
```diff
@@ -309,6 +309,42 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
+	}
+	return true
+}
+
+// rightBranchesAreEmpty returns true if the padding bytes correspond to all empty children
+// on the right side of this branch, ie. it's a valid placeholder on a rightmost path
+func rightBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
+	idx := getPosition(spec.ChildOrder, branch)
+	// compare the suffix bytes with the appropriate number of empty children
+	if len(op.Suffix) != idx*int(spec.ChildSize) {
```
ethanfrey, line 336: Again this is backwards.

`idx` is the number of left children, not the number of right children (items in suffix)

In this case, we want:

```go
idx := getPosition(spec.ChildOrder, branch)
rightBranches := len(spec.ChildOrder) - 1 - idx
if len(op.Suffix) != rightBranches*int(spec.ChildSize) {
  return false
}
for i := 0; i < rightBranches; i++ {
  // ...
}
```

ethanfrey (commented):
Thanks for implementing in both languages and providing these test cases.

Now that the test cases are a bit more readable, I can see where the test case "IsLeft"/"IsRight" does not match what I expect it to be. After digging in, I think there is a bug in the code. Can you please review/address this?

This is the same in both Go and Rust. With that addressed, I am happy to merge. 

I also noticed a missing feature. But we can talk about it later. The adjacency check doesn't handle if there are only emptyChild between X and Y. But let's talk later, the fixes mentioned above should be sufficient to merge

go/proof_test.go
```diff
@@ -54,3 +54,21 @@ func TestCheckAgainstSpec(t *testing.T) {
...
+				t.Errorf("Expected leftBranchesAreEmpty to be %t but it wasn't", tc.IsLeft)
+			}
+			if rightBranchesAreEmpty(tc.Spec.InnerSpec, tc.Op, 1) != tc.IsRight {
```
ethanfrey, line 69: Okay, I started to write a patch and see another error... you hardcode 0 and 1 as positions. Let's use the real position it has from padding.

ethanfrey:
Please review my comments. I made my suggested fixes to the Go side and it works. Or at least seems like it works for me.
There is some broken test in vectors as some previously invalid proof becomes valid... it would be good to investigate, but not critical - I think that setup is a bit artificial anyway.

Can you fix up the Rust side to match the Go side? Then we can merge this (comment out the vector test if you don't see what changed).

As a followup PR, `IsLeftNeighbor` also needs to be aware of such EmptyChild.

I assume you have 16-wide innernodes for the SMT, so we may very well set idx3 and idx6 but 4 and 5 are empty and yet those paths are adjacent. The current code works where `spec.ChildSize == 2` but not for all cases when it is larger.

go/proof_test.go
roysc: I think this is an error of naming and intended meaning. The field should probably have been `IsLeftPlaceholder`, since I meant for this to only be true for ops that contain an empty subtree on the left side.

It didn't make sense to me to call `leftBranchesAreEmpty` on the InnerOp describing a left branch, since by definition its left branch is non empty. So, I mistakenly used the branch index to instead get the padding for the branch on the opposite side (which of course works only when there are 2 children. I made it worse by hard coding the branch to make it work).

Even with your fix though, I would argue that these functions should not pass on Ops that describe a non-empty branch on the corresponding side. They only pass trivially now because the padding is empty, which I'd consider a bug.
So, I propose renaming the field to `IsLeftPlaceholder`, and changing the behavior so the functions fail when there is no padding to check for the corresponding side so e.g. `leftBranchesAreEmpty(..., branch=0)` is always false.

roysc:
Thanks for the fix, and for straightening me out here! I think I got confused by how InnerOp describes the _sibling_ nodes of a branch, and starting reversing indices in a misguided workaround for my confusion.


Let me finish the fixes and send it back to you tomorrow.

btw. We have a 2-wide inner node in this SMT implementation, but it would be interesting to test 3, 4, and 16, and I could look into that later.

go/proof.go
roysc: Yeah, this makes more sense. I was thinking about this backwards, such that `branch` meant, "the index over here" rather than, "the index corresponding to sibling data with this shape"

ethanfrey:
Okay, with this fix, this is correct for 2-wide.

We can make another PR later to ensure it works for 16-wide (needed for parity/substrate inter work) @seunlanlege

roysc:
Ready for another look now. I will try to start on the neighbor check fix and wider node support soon, could be interesting.

go/proof.go
```diff
@@ -313,13 +318,17 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
 // on the left side of this branch, ie. it's a valid placeholder on a leftmost path
 func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp, branch int32) bool {
 	idx := getPosition(spec.ChildOrder, branch)
-	// compare the prefix bytes with the appropriate number of empty children
-	leftChildren := idx
-	actualPrefix := len(op.Prefix) - leftChildren*int(spec.ChildSize)
+	// count branches to left of this
+	leftBranches := idx
+	if leftBranches == 0 {
+		return false
```
ethanfrey, line 324: That's an odd choice.
If there are no left branches, then all left branches are empty, right?
Why return false here?

(The true case would fall out after the loop anyway, but fine to explicitly return early. Not sure why you want to return false here: `on the left side of this branch, ie. it's a valid placeholder on a leftmost path` implies that no leftBranches at all would satisfy this)

rust/src/verify.rs
```diff
@@ -270,13 +270,18 @@ fn left_branches_are_empty(
     branch: i32,
 ) -> Result<bool> {
     if let Some(&idx) = spec.child_order.iter().find(|&&x| x == branch) {
-        // compare the prefix bytes with the appropriate number of empty children
-        let left_children = spec.child_order.len() - 1 - idx as usize;
+        // count branches to left of this
+        let left_branches = idx as usize;
+        if left_branches == 0 {
+            return Ok(false);
```
ethanfrey, line 276: I disagree with this for the same reason I disagree with the changes to the Go logic above.

left_branches == 0 not only satisfies what I think the function should mean but also the function comment in the Go code.

rust/src/verify.rs
```diff
@@ -270,13 +270,18 @@ fn left_branches_are_empty(
     branch: i32,
 ) -> Result<bool> {
     if let Some(&idx) = spec.child_order.iter().find(|&&x| x == branch) {
-        // compare the prefix bytes with the appropriate number of empty children
-        let left_children = spec.child_order.len() - 1 - idx as usize;
+        // count branches to left of this
+        let left_branches = idx as usize;
```
ethanfrey, line 274: Thanks for updating this part

ethanfrey (commented):
Thanks for updating the Rust logic as well.

I am very confused why you return `false` when there are no branches.

```
// leftBranchesAreEmpty returns true if the padding bytes correspond to all empty children
// on the left side of this branch, ie. it's a valid placeholder on a leftmost path
```

We can argue of the semantics of all on an empty set, but `it's a valid placeholder on a leftmost path` signals to me that we are looking for leftmost non-empty, which is true here

ethanfrey:
> Ready for another look now. I will try to start on the neighbor check fix and wider node support soon, could be interesting.

Yeah, let's do that in another PR.
I'd be happy to get this in soon, and https://github.com/confio/ics23/pull/57 for full Go support to validate SMT proofs, which I could get into IBC Go v3 hopefully.

I will follow up on https://github.com/confio/ics23/pull/62 but that is less critical as it requires a client sided update not a blockchain update (which takes longer to roll out)

go/proof.go
roysc: When the index is 0, the op describes a left-branch, but the op contains the data of the siblings of that branch, right? So `leftBranches` is the count of branches to the left of the "described" branch, which is none. It's not really a valid input to the function, because it asks a meaningless question: "are the branches to the left of this one empty?" That's how I see it, so an error may be more appropriate, but returning false seemed better than true.

roysc:
> We can argue of the semantics of all on an empty set, but `it's a valid placeholder on a leftmost path` signals to me that we are looking for leftmost non-empty, which is true here

Right, this function should be defined for ops that describe branches with valid left siblings (empty or not), and checks if those siblings are empty. An op that describe a leftmost branch has no left siblings, so it shouldn't be valid input anyway IMO. I will defer to you on the final behavior - if we go with true, I will look at why it breaks that vector test case.

I wonder if `leftSiblingsAreEmpty` might reflect the meaning more clearly?

ethanfrey:
> if we go with true, I will look at why it breaks that vector test case.

Returning false here fixes the vector test case? Weird. But a good hint.

I am off next two days, can look towards the end of the week

roysc:
@ethanfrey hey, do you expect to be able to review this week?

go/proof_test.go
ethanfrey: Since you will be the largest user if this, I can accept it. I am still not convinced this is the best or most maintainable way to write it.

I would like the function to say "isLeftmost" regardless of empty children to the left and **replace** the current padding check with just this function. Rather than check if it is actually the leftmost, and then check if it is not leftmost, but all (non-zero) left siblings are empty.

I think doing this logic in one piece would avoid confusion later. But I won't block this pr more.

go/proof.go
ethanfrey: again, as this function is only used to check "leftmost" of a path, I would do this differently. But no need to repeat myself

ethanfrey (approved):
merging with a few reservations.

the code is correct. we need to figure out some internal semantics.

PR #66, opened 2022-01-26
Use @noble/hashes

webmaster128:
This gives us lighter hash implementations.

See https://medium.com/nomic-labs-blog/a-safer-smaller-and-faster-ethereum-cryptography-stack-5eeb47f62d79 for the background of this implementation.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -32,8 +32,6 @@
   },
   "devDependencies": {
     "@types/jasmine": "^3.5.0",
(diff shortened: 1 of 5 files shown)


ethanfrey (approved):
looks good.

seems rust broke somehow (deps updated) and we need to bump the rust version used in the ci.

Care to make that PR? (here or separate)

PR #69, opened 2022-01-29
Update shx

webmaster128:
This is an alternative fix to #64 which is a bit more straight forward and updates not only shelljs


ethanfrey (approved):
thanks

PR #72, opened 2022-01-31
Migrate from tslint to eslint

webmaster128:
tslint is not maintained anymore. This uses more or less the eslint setup of CosmJS.


ethanfrey (approved):
lgtm

PR #74, opened 2022-02-01
Update TypeScript to 4.5 and target to es2017

webmaster128:
es2017 should easily be supported by all potential users. We will now migrate from es2017 to es2018 in CosmJS but we can take it easy here.

The TS upgrade does not seem to cause any trouble, nice.


ethanfrey (approved):
lgtm

PR #79, opened 2022-02-28
rust: add GH Actions CI

hdevalence:
From #78 it didn't seem like there was any Rust CI yet.


ethanfrey:
This is totally CI - Circle CI.. I guess it just doesn't run in forks

PR #88, opened 2022-05-17
Add new parameters to compare the given key and value for SMT proof

### LINKED ISSUE #83: Key comparison in proof generated from SMT
I'm wondering if the hashed key in proof generated from SMT can be compared in ibc-go's verification.
https://github.com/confio/ics23/blob/8ea0b239740068dd004fe4a42a1d861cd607611c/go/ics23.go#L138

In ibc-go, the proof is verified with the corresponding non-hashed key. However, the proof generated from SMT has only the hashed key. The verification would fail due to the key mismatching.

- Set the hashed key to an existence proof generated from SMT
https://github.com/cosmos/cosmos-sdk/blob/1ebe76c4801022c688613012f75f5406cc53e83d/store/tools/ics23/smt/create.go#L54

- Verify an existence proof in ibc-go
https://github.com/cosmos/ibc-go/blob/e59a6d2b88fbd543016b0a148b19fac368696ad9/modules/core/23-commitment/types/merkle.go#L247

Is that just because ibc-go has not supported SMT's proof yet?
I guess that ibc-go cannot avoid this issue because a chain trying to verify a proof doesn't know if the key in the proof is hashed or not.

yito88:
closes #83 

I'd like to add new parameters to the proof spec for SMT proof to check if the given key and value should be hashed before the comparison in the validations.
When a proof generated by SMT has the hashed key and value, the verification failed now due to the comparisons without hashing. That's why we want to modify the comparison by hashing the given key and value according to the new parameters in the proof spec.

I welcome discussion of the change or another suggestion for the proof verification for SMT proof.


ValarDragon (approved):
The change looks good to me! (Only 60 lines of go! Thats an insane amount of codegen happening, darn yarn.lock file lol)

Can you add a test vector demonstrating using the no-op at other layers of the tree, so we can have an in-golang test of correctness?

yito88:
Thank you for reviewing. I've moved the key checking and removed redundant code.
I'll add test vectors with non-hashed keys for SMT.

ethanfrey:
I looked very briefly.

This does change the semantics of how the library works and would be a breaking change, needed to be rolled out on all counterparty chains until such proofs could be used, right?

Would it not be possible to modify the calling code or formulate that on the  ibc-go side?
I would love input from the core ibc-go and SMT teams here before making any changes to this library

AdityaSripal:
>  I'd like to add new parameters to the proof spec for SMT proof to check if the given key and value should be hashed before the comparison in the validations.

Yes, I'd prefer to do the key hashing inside of the proof verification here rather than in ics23.

Preferably ibc-go interacts with all proofs the same way. It passes in an unhashed key, the unhashed value, and the proofspecs

Confio/ics23 looks at the proof spec, and determines if it needs to do any hashing before verification.

This would be the ideal case, because then any details about proof construction/verification are neatly encapsulated in the proof spec and the logic in this repo.

If it is not possible, then i suppose we could conditionally hash the key in ibc-go before passing it to confio but I would prefer not to do this.

ethanfrey:
If I understand it, you approve this approach @AdityaSripal ?

We will have to do this in a backwards compatible manner and ensure all counterparty chains have this before any chain updates to use SMT. I guess the best would get this in an confio/ics23 release soon and then include that in ibc-go v4, which can then roll out to all the chains?

ethanfrey:
I have a hard time tracking all downstream dependencies and it would be great if someone from IG can manage the various IBC implementations, so I know what is essential for the v0.8 release

AdityaSripal:
> before any chain updates to use SMT.

as I understand it, it was already the case that all counterparties have to upgrade anyway before SMT can be rolled out because the current versions cannot natively support an SMT tree. I don't believe there's any way of getting around this regardless of if the changes go in ibc-go or here?

> you approve this approach @AdityaSripal 

Yes I do. It seems to me if these changes can get in then there aren't any changes necessary to ibc-go code or relayer code.

I suppose the structures of the proofSpecs will change. Perhaps we can do something where if you are passed in a legacy proofSpec then you interpret the key hashing as a no-op (as it currently does)? This would then require no changes at all from my understanding to downstream dependencies. If we can't do that, then we may have to add some migration code in ibc-go to migrate all existing proof specs to add the new parameters

tagging @crodriguezvega to help provide context on any other considerations and syncing releases schedules

ethanfrey:
> Perhaps we can do something where if you are passed in a legacy proofSpec then you interpret the key/value hashing as a no-op (as it currently does)? 

Okay, let's make sure the zero-value for those fields have the same functionality as now, so they are interpreted in a backwards compatible manner.

Otherwise, waiting for passing test vectors to review and eventually approve

yito88:
Thank you for the discussion. I've rebased to resolve the conflicts.
CI failed due to `funty`. It failed with Rust 1.51.0. In my env. with 1.60.0, it passed.

yito88:
@ethanfrey I rebased it and all tests passed with new test vectors.

hdevalence:
> > Perhaps we can do something where if you are passed in a legacy proofSpec then you interpret the key/value hashing as a no-op (as it currently does)?
>
> Okay, let's make sure the zero-value for those fields have the same functionality as now, so they are interpreted in a backwards compatible manner.

I think this is already the case? If the `prehash_compared_key` or `prehash_compared_value` fields are missing, they'll be filled in with a default `HashOp`, and the zero value there is `NO_HASH` https://github.com/confio/ics23/blob/master/proofs.proto#L5-L7 .

So, this shouldn't be a breaking change to ICS23, it should already be backwards-compatible.

ethanfrey:
FYI, there is new ownership of this repo, it should be moved to cosmos org.

@AdityaSripal and @romac should be able to give direction (and others from their team)

go/ics23.go
```diff
@@ -36,11 +36,19 @@ type CommitmentRoot []byte
 func VerifyMembership(spec *ProofSpec, root CommitmentRoot, proof *CommitmentProof, key []byte, value []byte) bool {
 	// decompress it before running code (no-op if not compressed)
 	proof = Decompress(proof)
-	ep := getExistProofForKey(proof, key)
+	comparedKey, err := DoHashOrNoop(spec.GetPrehashComparedKey(), key)
+	if err != nil {
```
colin-axner, line 40: What about renaming `key` to `preimageKey` and then renaming `comparedKey` to `key` or `keyInTree`? 

From my understanding, the key passed in, is a preimage of the key used in the actual tree. This addition will then allow the tree to hash the preimage if it so chooses. I find it easier to reason about this way. Thoughts?

go/ics23.go
```diff
@@ -36,11 +36,19 @@ type CommitmentRoot []byte
...
+		return false
+	}
+	comparedValue, err := DoHashOrNoop(spec.GetPrehashComparedValue(), value)
+	if err != nil {
+		return false
```
colin-axner, line 45: Can you explain a use case of hashing the preimage value? Seems like it wouldn't be very practical as that means the tree doesn't store the preimage of the value and since hash functions are non-reversible, it'd need to be stored somewhere else anyways for an application to make use of the value. Thoughts?

go/ops.go
```diff
@@ -91,25 +91,25 @@ func (op *InnerOp) CheckAgainstSpec(spec *ProofSpec) error {
...
 	}
 	ldata, err := doLengthOp(lengthOp, hdata)
 	return ldata, err
 }
 
-// doHashOrNoop will return the preimage untouched if hashOp == NONE,
+// DoHashOrNoop will return the preimage untouched if hashOp == NONE,
 // otherwise, perform doHash
-func doHashOrNoop(hashOp HashOp, preimage []byte) ([]byte, error) {
+func DoHashOrNoop(hashOp HashOp, preimage []byte) ([]byte, error) {
```
colin-axner, line 104: If possible, I think it makes sense to keep this function private. Was there a reason to make it public?

go/proof.go
```diff
@@ -58,6 +58,7 @@ var SmtSpec = &ProofSpec{
 		Hash:            HashOp_SHA256,
 	},
 	MaxDepth: 256,
+  PrehashComparedKey: HashOp_SHA256,
```
colin-axner, line 61: I find the naming between `PrehashKey` and `PrehashComparedKey` confusing. Thanks @AdityaSripal for helping me walkthrough the difference. From my understanding:

Fields in the `LeafOp` represent how we encode the Leaf node when constructing and proving proofs. The `PrehashComparedKey` is used to construct the key used in the tree (ie its location). 

Are there ways we could improve the field naming here such that its clear that these fields are distinctly different despite both being prehashes on a key. In my terminology, I see `PrehashComparedKey` as a prehash on the preimage of the key, while `PrehashKey` is a prehash of the key. Maybe something like `PrehashPreimageKey`?

go/proof.go
```diff
@@ -92,18 +93,11 @@ func (p *CommitmentProof) Calculate() (CommitmentRoot, error) {
 
 // Verify does all checks to ensure this proof proves this key, value -> root
 // and matches the spec.
-func (p *ExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []byte, value []byte) error {
+func (p *ExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot) error {
```
colin-axner, line 96: Would it be possible to make this private? I think it'd be good if ics23 only exposed `VerifyMembership` and `VerifyNonMembership` as it's two public APIs

colin-axner (commented):
Thanks @yito88! I left only code improvement/readability suggestions. I only reviewed the go code, but conceptually the changes look great to me. 

For reference, when I say "preimage key", I am referring to a key ibc-go might have labeled as "ibc/channels/channel-0", that is, the key used by the application, whereas to me "key" refers to the actual key stored in the SMT or IAVL tree. It seems the approach decided on is to continue expecting applications such as ibc-go to pass in the preimage key and have ics23 do all the underlying tree construction logic (turning the preimage key into the actual key in the tree). Very open to alternative naming conventions, but would be nice to have something clear and concise and easy to understand. Perhaps (applicationKey and treeKey or something along those lines could work as well)

go/ics23.go
AdityaSripal: It seems to me we should just leave this out and revisit once there is an actual user for this field

yito88: In our case, the tree (Sparse Merkle Tree) is stored on memory and has hashed values with a fixed length. It only checks the brief existence (checks whether we need to read the original value on the disk) and gets the proofs.

yito88:
@colin-axner @AdityaSripal Thank you for reviewing this PR! I'll restart this according to your feedback.

hdevalence:
Just an FYI, after thinking through how we'd use this for Penumbra, we think this approach may not be the best way to extend the existing ICS23 proof specs to support sparse merkle trees, and are finishing up an alternative proposal that has the same backward compatibility properties but is more general. We'll have a PR for it in the coming days.

avahowell:
Our alternative proposal is here: https://github.com/cosmos/ics23/pull/136

crodriguezvega:
Closing this now that https://github.com/cosmos/ics23/pull/136 has been merged.

PR #90, opened 2022-06-03
adds support for host functions to ics23 

seunlanlege:
This encapsulates all crypto hash functions behind a trait, so downstream crates can delegate these functions to host functions

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -18,6 +18,10 @@ bytes = { version = "1.0.1", default-features = false }
 hex = { version = "0.4.3", default-features = false, features = [ "alloc" ] }
 anyhow = { version = "1.0.40", default-features = false }
 sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false }
+sha2 = { version = "0.9.3", optional = true }
+sha3 = { version = "0.9.1", optional = true }
+ripemd160 = { version = "0.9.1", optional = true }
+sp-core = { version = "6.0.0", optional = true }
 
 [dev-dependencies]
 sha2 = { version = "0.9.3" }
@@ -35,4 +39,8 @@ std = [
     "hex/std",
     "anyhow/std",
     "sp-std/std",
+    "sha2",
+    "sha3",
+    "ripemd160",
+    "sp-core",
 ]
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -16,13 +16,14 @@ members = ["codegen"]
 prost = { version = "0.10", default-features = false, features = ["prost-derive"] }
 bytes = { version = "1.0.1", default-features = false }
 hex = { version = "0.4.3", default-features = false, features = [ "alloc" ] }
-sha2 = { version = "0.9.3", default-features = false }
-sha3 = { version = "0.9.1", default-features = false }
-ripemd160 = { version = "0.9.1", default-features = false }
 anyhow = { version = "1.0.40", default-features = false }
-sp-std = {version = "3.0.0", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false }
 
 [dev-dependencies]
+sha2 = { version = "0.9.3" }
+sha3 = { version = "0.9.1" }
+ripemd160 = { version = "0.9.1" }
+sp-core = { version = "6.0.0" }
 serde = { version = "1.0.125", features = ["derive"] }
 serde_json = "1.0.64"
 
@@ -32,9 +33,6 @@ std = [
     "prost/std",
     "bytes/std",
     "hex/std",
-    "sha2/std",
-    "sha3/std",
-    "ripemd160/std",
     "anyhow/std",
     "sp-std/std",
 ]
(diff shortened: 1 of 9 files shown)


ethanfrey (commented):
Thanks for your PR. I agree with the general direction, but request some design changes.

Rather than keep the `HostFunctionsManager` only in test code (as dev-dependency), I would rather hide it under a feature flag. And ideally have that flag on by default.

For substrate, you can then use `no-default-features` to avoid this dependency and use the host functions. But it would be a lot less intrusive for other projects (like ibc-rs) that also run in the Hermes relayer and want such functions as well. They should not have to implement this themselves.

rust/Cargo.toml
```diff
@@ -16,13 +16,14 @@ members = ["codegen"]
 prost = { version = "0.10", default-features = false, features = ["prost-derive"] }
 bytes = { version = "1.0.1", default-features = false }
 hex = { version = "0.4.3", default-features = false, features = [ "alloc" ] }
-sha2 = { version = "0.9.3", default-features = false }
-sha3 = { version = "0.9.1", default-features = false }
-ripemd160 = { version = "0.9.1", default-features = false }
 anyhow = { version = "1.0.40", default-features = false }
-sp-std = {version = "3.0.0", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false }
```
ethanfrey, line 20: I really don't like tagging to branches here.
Can we just update to a newer release if that is what you need?

seunlanlege: unfortunately, this won't work. Substrate doesn't publish new releases to crates.io so parachain teams need to track git branches for new releases.

rust/src/host_functions.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[cfg(test)]
+pub mod test_helper {
+    use crate::host_functions::HostFunctionsProvider;
+    use ripemd160::Ripemd160;
+    use sha2::{Digest, Sha512, Sha512Trunc256};
+    use sha3::Sha3_512;
+
+    pub struct HostFunctionsManager;
```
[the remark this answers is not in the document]
seunlanlege: Its available in std now

rust/Cargo.toml
ethanfrey: why do they do that? do you know?

seunlanlege: from the versioning (currently we're at 0.9.24) I'd say they don't want to publish until they hit v1.0.0

rust/src/lib.rs
```diff
@@ -21,3 +20,4 @@ pub use api::{
 pub use compress::{compress, decompress, is_compressed};
 pub use helpers::{Hash, Result};
 pub use verify::calculate_existence_root;
+pub use host_functions::HostFunctionsProvider;
```
ethanfrey, line 23: Please add

```
#[cfg(feature = "std")]
pub use host_functions::host_functions_impl::HostFunctionsManager;
```

rust/src/ops.rs
```diff
@@ -221,7 +224,7 @@ mod tests {
         // echo -n 012345678900cafe00deadbeef | xxd -r -p | sha256sum
         let expected = decode("0339f76086684506a6d42a60da4b5a719febd4d96d8b8d85ae92849e3a849a5e");
         assert!(
-            expected == apply_inner(&inner, &child).unwrap(),
+            expected == apply_inner::<HostFunctionsManager>(&inner, &child).unwrap(),
```
ethanfrey, line 227: It would be good to leave some CHANGELOG entry explaining the API change (use `xyz::<HostFunctionsManager>` rather than `xyz`) that work for the general std use cases. I know `ibc-rs` and penumbra are using this as well in std mode.

ethanfrey (commented):
Looks good, one last comment for minor updates

ethanfrey:
@hdevalence @adizere @romanc I am pretty sure you all use the std variant here. This change should be compatible, just allow more flexibility at the cost of requiring [specifying a hash implementation](https://github.com/confio/ics23/pull/90/files/c42fb738f19e6cdfca07fa7ffb5766dde6368feb#diff-3aa19cf462646004beef91440a07cef152db4d15a36c9c656a2ea32cdc61af4c) when calling some functions.

I will likely merge this shortly, but I would really like your feedback before tagging anything, we can still adjust the API to work for all

rust/Cargo.toml
adizere: @ethanfrey will you be able to tag a version & [publish on crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) with the dependency specified as git/branch?

romanc:
> @hdevalence @adizere @romanc I am pretty sure you all use the std variant here. This change should be compatible, just allow more flexibility at the cost of requiring [specifying a hash implementation](https://github.com/confio/ics23/pull/90/files/c42fb738f19e6cdfca07fa7ffb5766dde6368feb#diff-3aa19cf462646004beef91440a07cef152db4d15a36c9c656a2ea32cdc61af4c) when calling some functions.
> 
> I will likely merge this shortly, but I would really like your feedback before tagging anything, we can still adjust the API to work for all

@ethanfrey I think you tagged me by mistake. I am not part of your project. Just letting you know because it sounded like you wanted feedback from someone else.

PR #107, opened 2022-10-07
Fix link to spec

ebuchman:
Fix link to ics23 spec under ibc/core

### DIFF
--- a/go/ics23.go
+++ b/go/ics23.go
@@ -1,6 +1,6 @@
 /**
 This implements the client side functions as specified in
-https://github.com/cosmos/ics/tree/master/spec/ics-023-vector-commitments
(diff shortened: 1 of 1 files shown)


go/ics23.go
```diff
@@ -1,6 +1,6 @@
 /**
 This implements the client side functions as specified in
-https://github.com/cosmos/ics/tree/master/spec/ics-023-vector-commitments
+https://github.com/cosmos/ibc/tree/master/spec/core/ics-023-vector-commitments
```
plafer, line 3:
```suggestion
https://github.com/cosmos/ibc/tree/main/spec/core/ics-023-vector-commitments
```
Might as well update the branch name too while we're at it

plafer (approved):

PR #131, opened 2022-12-19
chore: retract previous versions

### LINKED ISSUE #125: Retract ics23 0.7
Now that there is 0.9 with the fix, should not we retract 0.7 in the `go.mod`?
https://github.com/cosmos/ics23/blob/master/go/go.mod

```go
// dragonberry vulnerability
retract v0.7.0
```

julienrbrt:
Closes #125

This will be enforced at next ICS23 tag.

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -6,3 +6,6 @@ require (
 	github.com/cosmos/gogoproto v1.4.3
 	golang.org/x/crypto v0.2.0
 )
+
+// subject to the dragonberry vulnerability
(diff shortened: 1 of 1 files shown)


tac0turtle (commented):
We should see if this is breaking for ibc versions

julienrbrt:
> We should see if this is breaking for ibc versions 

I don't think it's breaking as retracting a version does not prohibit it to be downloaded if already imported by a package.

tac0turtle (approved):

PR #132, opened 2023-01-03
support BLAKE3 cryptographic hash function

shuimuliang:


shuimuliang:
support BLAKE3 cryptographic hash function
https://github.com/BLAKE3-team/BLAKE3

tac0turtle:
@colin-axner do we have to support this in go as well?

PR #135, opened 2023-02-21
feat: support BLAKE3 cryptographic hash function 

shuimuliang:
according to the comment:
1. restore Cargo.toml
name = "ics23"
repository = "https://github.com/cosmos/ics23/tree/master/rust"

2. test case fixed in rust/src/ops.rs
`assert!(hash == decode("f775a8ccf8cb78cd1c63ade4e9802de4ead836b36cea35242accf31d2c6a3697"), "blake3 hash fails");`

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -21,6 +21,7 @@ prost = {version = "0.11", default-features = false, features = ["prost-derive"]
 ripemd = {version = "0.1.1", optional = true, default-features = false}
 sha2 = {version = "0.10.2", optional = true, default-features = false}
 sha3 = {version = "0.10.2", optional = true, default-features = false}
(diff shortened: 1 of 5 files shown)


hdevalence:
If ICS23 is going to be extended with BLAKE, it'd be helpful to add support for Blake2s and Blake2b. We'd actually rather have used Blake2 for the Penumbra JMT to match the rest of our crypto, but decided to stick with SHA2 for compatibility with other IBC chains, because it wasn't supported by ICS23.

crodriguezvega:
@shuimuliang Sorry for the late reply... Thanks for the PR, but we will move forward with https://github.com/cosmos/ics23/pull/212 instead to add the Blake hashing functions. Apologies for that!

PR #136, opened 2023-02-22
feat: Add `prehash_compare_key` to allow proving nonexistence in sparse trees

plaidfinch:
## Motivation

There are several important merkle trees that hash keys and sparsely store values at positions indicated by the value of the hash, such as [the Cosmos Sparse Merkle Tree](https://github.com/cosmos/cosmos-sdk/tree/1ebe76c4801022c688613012f75f5406cc53e83d/store/tools/ics23/smt) and [Libra/Diem/Aptos/Penumbra's Jellyfish Merkle Tree](https://github.com/penumbra-zone/jmt). We would like this type of tree to be compatible with ICS23.

For existence proofs, this is already the case; however, currently, when verifying a ICS23 _non_-existence proof, keys are compared based on the preimage of the hash function used to prehash them, **even when `prehash_key` is set in the `ProofSpec`**. This means that when a `ProofSpec` uses `prehash_key`, nonexistence proofs cannot be verified even if a client correctly generates them as per the "spirit" of the ICS23 nonexistence proof format. We believe that this is a bug, but this change is framed backwards-compatibly, as noted below, so that it can be applied universally without friction.

See here for other discussion of the impact of this issue:

- https://github.com/penumbra-zone/jmt/issues/24#issuecomment-1248232171 (Note: We think that ibc-go can't make any assumption about the keys being hashed or unhashed, since that's not part of its domain; that's the domain of the proof spec. However, this description of the issue is pointing at the same problem that this PR resolves.)
- https://github.com/cosmos/ics23/issues/83 is the corresponding issue for the SMT (Note: In the [current SMT implementation](https://github.com/cosmos/cosmos-sdk/blob/1ebe76c4801022c688613012f75f5406cc53e83d/store/tools/ics23/smt/create.go#L54), keys are prehashed entirely externally to the proof and its spec, and the relationship between key and hashed key is proven externally to the proof spec. We believe that this ends up adding implementation burden for counterparty chains, who have to check the additional constraint outside of ICS23 verification that the key being proven matches the hash claimed. Instead of this, the [JMT proof spec](https://github.com/penumbra-zone/jmt/blob/exclusion-proofs/src/tree/ics23_impl.rs#L222) could be adapted to the SMT, allowing the SMT to internalize the prehashing and avoid the difficulty of managing prehashing externally to the proof spec. @cwgoes: does this seem like something you would want to do?)

## Summary of changes

This PR introduces **one single boolean** new parameter to the top-level `ProofSpec`: `prehash_compare_key`. When set to `true`, this flag causes keys to be consistently compared lexicographically according to their _hashes_ within nonexistence proof verification, using the same hash function as specified by the already-extant `prehash_key` field.

## This is intended as an alternative to #88

We believe this change is the minimum necessary to unblock nonexistence proofs on key-prehashed structures. #88 also attempts to solve this problem, namely that for _nonexistence proofs only_, any tree that uses prehashing cannot prove nonexistence because the nonexistence proof verifier compares the keys lexically by their preimage, ignoring the `prehash_key` field of the specification.

However, as currently implemented, **#88 does not accomplish this goal.** There are 3 issues with #88 as a solution to this problem, which this PR addresses:

1. It does not actually change the verification procedure for nonexistence proofs to lexically compare all keys by their hash; rather, it [lexically compares only the hash of the input key to unhashed neighbor keys](https://github.com/cosmos/ics23/commit/473d9a546e5cbc5c645f700b83ceec3f50645ccd#diff-3f131e8084f824850331466f4c866266af1532478845d7567db58495d7654a02R219). **This will not work correctly for any general cryptographic hash function**, because comparing a cryptographic hash to an element of its preimage is, effectively, random.
2. It doesn't merely supply a flag opting into prehashed key comparison; it allows you to specify an entirely different hash function to use for the comparison. We believe this is not representative of any known use case: if keys are prehashed using hash function H, then a different hash function H' can't be used to compute a meaningful comparison on keys, so this is more complex of a specification than necessary.
3. It also introduces a `prehash_compared_value` field, which is not necessary to fix this specific issue, and like its `prehash_compared_key` field, this too is a specification of an arbitrary hash function, which for the same reasons above, we believe is overly general.

## Backwards-compatibility

This is a backwards-compatible change, as it requires opt-in via setting the `prehash_compare_key` flag to `true` in the `ProofSpec`. All existing `ProofSpec`s will continue to behave identically.

## Contents

This PR includes implementation for Rust, Go, and Typescript. We have [tested it against our own implementation of the Jellyfish Merkle Tree](https://github.com/penumbra-zone/jmt/blob/4a96055cbc70f3baef5bd73978d332880c601ded/src/tree/ics23_impl.rs#L251) and we believe it should be effective at addressing this issue across the ecosystem.

## Thanks

We would like to thank everyone who participated in discussion and development of #88 and other work towards solutions to this issue. We feel very confident that this is the right way forward, but we want to ensure that our contribution does not make anyone feel like their work is unappreciated; to the contrary, the discussion and work leading up to this PR, by everyone, has been necessary to clarify our understanding of the issue. Thanks everyone for your help!

### DIFF
--- a/js/src/testvectors.spec.ts
+++ b/js/src/testvectors.spec.ts
@@ -8,7 +8,7 @@ import {
   verifyMembership,
   verifyNonMembership,
 } from "./ics23";
-import { iavlSpec, tendermintSpec } from "./proofs";
+import { iavlSpec, smtSpec, tendermintSpec } from "./proofs";
 import { fromHex } from "./testhelpers.spec";
 
 describe("calculateExistenceRoot", () => {
@@ -250,5 +250,29 @@ describe("calculateExistenceRoot", () => {
       "../testdata/tendermint/nonexist_middle.json",
     ]);
     validateBatch(proof, tendermintSpec, data[3]);
+  });
+
+  it("should validate smt batch exist", () => {
+    const { proof, data } = loadBatch([
+      "../testdata/smt/exist_left.json",
+      "../testdata/smt/exist_right.json",
+      "../testdata/smt/exist_middle.json",
+      "../testdata/smt/nonexist_left.json",
+      "../testdata/smt/nonexist_right.json",
+      "../testdata/smt/nonexist_middle.json",
+    ]);
+    validateBatch(proof, smtSpec, data[2]);
+  });
+
+  it("should validate smt batch nonexist", () => {
+    const { proof, data } = loadBatch([
+      "../testdata/smt/exist_left.json",
+      "../testdata/smt/exist_right.json",
+      "../testdata/smt/exist_middle.json",
+      "../testdata/smt/nonexist_left.json",
+      "../testdata/smt/nonexist_right.json",
+      "../testdata/smt/nonexist_middle.json",
+    ]);
+    validateBatch(proof, smtSpec, data[3]);
   });
 });
--- a/js/yarn.lock
+++ b/js/yarn.lock
@@ -301,7 +301,7 @@
     "fastq" "^1.6.0"
 
 "@protobufjs/aspromise@^1.1.1", "@protobufjs/aspromise@^1.1.2":
-  "integrity" "sha1-m4sMxmPWaafY9vXQiToU00jzD78= sha512-j+gKExEuLmKwvz3OgROXtrJ2UG2x8Ch2YZUxahh+s1F2HZ+wAceUNLkvy6zKCPVRkU++ZWQrdxsUeQXmcg4uoQ=="
+  "integrity" "sha1-m4sMxmPWaafY9vXQiToU00jzD78=sha512-j+gKExEuLmKwvz3OgROXtrJ2UG2x8Ch2YZUxahh+s1F2HZ+wAceUNLkvy6zKCPVRkU++ZWQrdxsUeQXmcg4uoQ== sha512-j+gKExEuLmKwvz3OgROXtrJ2UG2x8Ch2YZUxahh+s1F2HZ+wAceUNLkvy6zKCPVRkU++ZWQrdxsUeQXmcg4uoQ=="
   "resolved" "https://registry.npmjs.org/@protobufjs/aspromise/-/aspromise-1.1.2.tgz"
   "version" "1.1.2"
 
@@ -316,12 +316,12 @@
   "version" "2.0.4"
 
 "@protobufjs/eventemitter@^1.1.0":
-  "integrity" "sha1-NVy8mLr61ZePntCV85diHx0Ga3A= sha512-j9ednRT81vYJ9OfVuXG6ERSTdEL1xVsNgqpkxMsbIabzSo3goCjDIveeGv5d03om39ML71RdmrGNjG5SReBP/Q=="
+  "integrity" "sha1-NVy8mLr61ZePntCV85diHx0Ga3A=sha512-j9ednRT81vYJ9OfVuXG6ERSTdEL1xVsNgqpkxMsbIabzSo3goCjDIveeGv5d03om39ML71RdmrGNjG5SReBP/Q== sha512-j9ednRT81vYJ9OfVuXG6ERSTdEL1xVsNgqpkxMsbIabzSo3goCjDIveeGv5d03om39ML71RdmrGNjG5SReBP/Q=="
   "resolved" "https://registry.npmjs.org/@protobufjs/eventemitter/-/eventemitter-1.1.0.tgz"
   "version" "1.1.0"
 
 "@protobufjs/fetch@^1.1.0":
-  "integrity" "sha1-upn7WYYUr2VwDBYZ/wbUVLDYTEU= sha512-lljVXpqXebpsijW71PZaCYeIcE5on1w5DlQy5WH6GLbFryLUrBD4932W/E2BSpfRJWseIL4v/KPgBFxDOIdKpQ=="
+  "integrity" "sha1-upn7WYYUr2VwDBYZ/wbUVLDYTEU=sha512-lljVXpqXebpsijW71PZaCYeIcE5on1w5DlQy5WH6GLbFryLUrBD4932W/E2BSpfRJWseIL4v/KPgBFxDOIdKpQ== sha512-lljVXpqXebpsijW71PZaCYeIcE5on1w5DlQy5WH6GLbFryLUrBD4932W/E2BSpfRJWseIL4v/KPgBFxDOIdKpQ=="
   "resolved" "https://registry.npmjs.org/@protobufjs/fetch/-/fetch-1.1.0.tgz"
   "version" "1.1.0"
   dependencies:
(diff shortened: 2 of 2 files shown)


AdityaSripal:
Thank you for including the justifications to include this change over #88, all of them make sense to me.

Will review today!

proto/cosmos/ics23/v1/proofs.proto
```diff
@@ -163,6 +163,10 @@ message ProofSpec {
   int32 max_depth = 3;
   // min_depth (if > 0) is the minimum number of InnerOps allowed (mainly for fixed-depth tries)
   int32 min_depth = 4;
+  // prehash_compare_key is a flag that indicates whether to use the
+  // prehash_key specified by LeafOp to compare lexical ordering of keys for
+  // non-existence proofs.
+  bool prehash_compare_key = 5;
```
AdityaSripal, line 169: I think a clearer name might be helpful here similar to our review of #88.

So effectively the database is providing an interface to the application: `store.Set(key, value)`

and underneath the hood it is merklizing this pair by storing the hash of the application-provided key as the key in the merkle tree.

Thus, we liked the nomenclature: `appKey` and `treeKey`.

The `appKey` is the key known to the application, the `treeKey` is the key stored in the tree.

In SMT, the treeKey is the hash of the appKey, while in iAVL they are the same.

So I think `prehash_app_key` would be clearer here as a name. And we can use the docs here to explain there may be a difference between appKey and treeKey for some  trees

go/proof.go
```diff
@@ -204,6 +204,14 @@ func (p *ExistenceProof) CheckAgainstSpec(spec *ProofSpec) error {
 	return nil
 }
 
+func keyForComparison(spec *ProofSpec, key []byte) []byte {
```
AdityaSripal, line 207: You also will have to use this function on `getExistProofForKey` and `getNonexistProofForKey` correct?

https://github.com/cosmos/ics23/commit/473d9a546e5cbc5c645f700b83ceec3f50645ccd#diff-39a55415fc38a90b85c05989f293fda3b7ee126010cfe63838fd9b8441e47ed1R39

AdityaSripal (changes requested):
Agreed with this direction!! But it looks like this PR is incomplete?

My understanding is that the key in `ExistenceProof.Key` will still be the preimage, hence why you also hash the rightKey and leftKey before comparison.

However, you are not hashing the passed-in key for existence proofs before comparison?

There is a check on ics23.go that is changed in #88 that seems like it would also need to be changed here.

https://github.com/cosmos/ics23/commit/473d9a546e5cbc5c645f700b83ceec3f50645ccd#diff-39a55415fc38a90b85c05989f293fda3b7ee126010cfe63838fd9b8441e47ed1R39

https://github.com/cosmos/ics23/commit/473d9a546e5cbc5c645f700b83ceec3f50645ccd#diff-39a55415fc38a90b85c05989f293fda3b7ee126010cfe63838fd9b8441e47ed1R59

I've also suggested a different naming for the field that at least me and Colin found more intuitive

proto/cosmos/ics23/v1/proofs.proto
hdevalence: This is a bit of a bikeshed, but if there's going to be a distinction between `app_key` and `tree_key`, wouldn't it be even clearer to call the field `compare_tree_key`?  Then `compare_tree_key = false` means it compares the app key, and `compare_tree_key = true` means it compares the tree key.

go/proof.go
avahowell: Yes, good catch

avahowell: but only on `getNonexistProof`, because the lexical comparison is only relevant for non-existence proofs

proto/cosmos/ics23/v1/proofs.proto
AdityaSripal: Ok now that I understand more of this field.

Perhaps we can call it: `prehash_key_before_comparison`

The way I read the current field name at the moment, I expect it to be a HashOp type not a boolean.

cc: @colin-axner

go/proof.go
```diff
@@ -204,6 +204,14 @@ func (p *ExistenceProof) CheckAgainstSpec(spec *ProofSpec) error {
...
+	if !spec.PrehashCompareKey {
+		return key
+	}
+	hash, _ := doHashOrNoop(spec.LeafSpec.PrehashKey, key)
```
AdityaSripal, line 211: Since this is using the PrehashKey defined in the leafspec and the key in the existenceProof is supposed to now be unhashed for SMT proofs, I think the SMT spec needs to be changed to have `PrehashKey=SHA_256` instead of `NO_HASH`

AdityaSripal (changes requested):

go/proof.go
avahowell: I think the SMT should mirror the JMT's proof spec:

https://github.com/penumbra-zone/jmt/blob/main/src/tree/ics23_impl.rs#L219

AdityaSripal: Right so the SMT proof spec in this repo need to be changed

avahowell: The code that generates the proofs will also need to be updated following the spec change

joebebel:
Backwards compatibility nonwithstanding, it seems like applications (if they can) should *always* set `prehash_key_before_comparison` to `true` (even in the absence of actual prehashing) because the `false` behavior is (as described) inconsistent/not intuitive and/or "buggy"; if so, perhaps this can be documented in `ProofSpec`.

cwgoes:
@plaidfinch Thanks for this! I can tell you guys have thought this through. I think that this solution should work for us as well.

avahowell:
> Backwards compatibility nonwithstanding, it seems like applications (if they can) should _always_ set `prehash_key_before_comparison` to `true` (even in the absence of actual prehashing) because the `false` behavior is (as described) inconsistent/not intuitive and/or "buggy"; if so, perhaps this can be documented in `ProofSpec`.

Yes, the only reason to have a flag in this case is to maintain strict backwards-compatibility in my opinion. If you have `prehash_key_before_comparison: true` on a tree with no prehashing, it's the same thing as having `prehash_key_before_comparison: false`, since this change just inherits the hashing specified by `prehash_key`. 

Where the change would be breaking without the flag is in the case where you use `prehash_key`, but don't want to compare keys according to their prehashing algorithm.

avahowell:
I've updated this code to include the changes for the `smt_spec` required for the SMT. The next step is to update the SMT's proof-generating code itself to be compatible with this new spec.

go/proof.go
```diff
@@ -204,6 +205,14 @@ func (p *ExistenceProof) CheckAgainstSpec(spec *ProofSpec) error {
 	return nil
 }
 
+func keyForComparison(spec *ProofSpec, key []byte) []byte {
```
colin-axner, line 208: might be nice to add a little doc string explaining the usage of this function, that is to obtain the key used for lexical ordering in non existence proofs because prehashing the key changes the lexical ordering.

go/ics23.go
```diff
@@ -148,27 +148,27 @@ func getExistProofForKey(proof *CommitmentProof, key []byte) *ExistenceProof {
 	return nil
 }
 
-func getNonExistProofForKey(proof *CommitmentProof, key []byte) *NonExistenceProof {
+func getNonExistProofForKey(spec *ProofSpec, proof *CommitmentProof, key []byte) *NonExistenceProof {
```
colin-axner, line 151: it would be nice if we could add a test case for the scenario that was previously failing

colin-axner (commented):
ACK for the concept and the go code changes. I like this solution a lot! Excellent work y'all :)

avahowell:
i updated the smt test vectors to be correctly generated per this change, and they should verify now. i generated the test vectors using a modified version of the smt in the `store/v2` branch of the cosmos sdk, which I can PR if desired. Here's the branch, the change required for the smt on the generation side is very small since it already stores a PreimageMap:

https://github.com/avahowell/cosmos-sdk/commit/a3a049bc2577632b94ff2236c2b01a250e85b16d

AdityaSripal:
Hi yes, please make a PR to the SDK to update the proofgen code there

colin-axner:
Would it be possible to eventually have test data vectors from Penumbra's JMT? It appears the referenced code generation has been removed from the SDK and I'm not sure the SMT implementation is being maintained

rust/codegen/src/main.rs
```diff
@@ -16,7 +16,6 @@ fn main() {
     prost_build::Config::new()
         .out_dir(&out_dir)
         .format(true)
-        .type_attribute(".", "#[derive(Eq)]")
```
romac, line 19: What's the reason for removing this?

rust/src/api.rs
```diff
@@ -135,18 +136,50 @@ fn get_exist_proof<'a>(
...
     proof: &'a ics23::CommitmentProof,
     key: &[u8],
+    spec: &ics23::ProofSpec,
 ) -> Option<&'a ics23::NonExistenceProof> {
+    // If we should prehash the key before comparison, do so; otherwise, return the key. Prehashing
+    // changes lexical comparison, so if the spec sets `prehash_key_before_comparison` to true, we
+    // should compare keys on their hashes by prehashing them before comparison.
+    let key_for_comparison = |key: &[u8]| match spec.prehash_key_before_comparison {
+        true => do_hash::<H>(
+            spec.leaf_spec.clone().unwrap_or_default().prehash_key(),
```
romac, line 149: nit: We could avoid cloning here by doing something like this, but it does not read as nice as the current version imho, which makes it clear we are using the default/empty `LeafSpec`. 

```suggestion
            spec.leaf_spec.as_ref().map(|leaf_spec| leaf_spec.prehash_key()).unwrap_or_default(),
            
```

romac (approved):
Rust version looks good! I just left one question regarding the removal of the derived `Eq` instance and a small nit, feel free to ignore.

colin-axner:
> Would you want these test vectors included in this PR before merging it, or should we make a separate PR with them? It'd require a bit of implementation work to add the code to the JMT to make it spit out the test vectors, so my personal preference would be to merge this PR and make a separate PR later to swap in the JMT test vectors for the SMT ones.

Let's do a separate pr :+1: I don't see a rush in adding the test vectors, just want to make sure that long term ics23 has an up to date testing framework/tests :smile:

rust/codegen/src/main.rs
romac: Let's revert this then if @avahowell doesn't object?

avahowell: I had to remove these due to them causing conflicting implementations and breaking the generated code on my machine (test with `cargo test` after running codegen):

```
error[E0119]: conflicting implementations of trait `std::cmp::Eq` for type `ics23::LengthOp`
```

romac (approved):

romac:
All good on the Rust side! I can do a release of the Rust crate on Tuesday if nobody beats me to it.

hdevalence:
Hey, just checking in on this -- are we still good to merge this PR and cut a release?

Olshansk:
Support for SMTs is going to go a long way. Thanks for everyone involved here!

PR #140, opened 2023-04-10
test: add benchmark tests for verify membership/non-membership

crodriguezvega:
This PR adds a set of benchmark tests for the verify membership/non-membership functions. This is an exploratory PR to gauge interest on having this.

The PR uses [this GitHub action](https://github.com/benchmark-action/github-action-benchmark) to add workflow to commit to a `gh-pages` branch the results of each benchmark run. The benchmark results can then be access on the browser. See example [here](https://crodriguezvega.github.io/ics23/dev/bench/) for the charts after 2 runs from a forked repository of ics23.

When a PR is open to ics23 the GitHub action will fail if benchmark results are worse than previous one by a configurable threshold and will write a comment on the commit that caused the degradation. See [here](https://github.com/crodriguezvega/ics23/commit/e968a1e6ee92633b63f610c8bacc151b5a13752e#comments) for an example for a PR to the forked repository of ics23 that simulated an increase of nanoseconds per operation. See here for an example of [the action that failed on the fork](https://github.com/crodriguezvega/ics23/actions/runs/4652064871/jobs/8232053049).

I hope this is found to be useful!


.github/workflows/go.yml
```diff
@@ -62,3 +63,26 @@ jobs:
...
+          name: Go Benchmark
+          tool: 'go'
+          output-file-path: ./go/output.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '50%'
+          comment-on-alert: true
+          fail-on-alert: true
+          alert-comment-cc-users: '@crodriguezvega'
```
crodriguezvega, line 88: These are the GitHub handles that will be mentioned in the comment when the benchmark results are worse comparing to the previous one.

go/benchmarks/benchmark_test.go
```diff
@@ -0,0 +1,183 @@
+package benchmarks
```
crodriguezvega, line 1: I took the [iavl benchmark test](https://github.com/cosmos/iavl/blob/master/benchmarks/bench_test.go) as base for this one.

crodriguezvega:
> is this still relevant? Lmk if you want me to make up some time to review.

I think it is. :) Let me fix the conflicts and I will ping you. Thank you in advance. :)

PR #142, opened 2023-04-12
chore: Prepare for v0.10.0 release

romac:


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+# 0.10.0
```
crodriguezvega, line 3: Just noticing now that if the JS code will be released as v0.7.0, should we split the changelog? Or should we just for simplicity bump JS also to v0.10.0 and then we keep all of the in sync?

romac: Yeah let's bump it to 0.10.0 as well. We may not be able to keep all three versions in sync forever if changes need to happen in only a single implementation but perhaps we'll find a versioning scheme that helps us keep at least the major/minor versions in sync.

crodriguezvega (approved):
Thank you, @romac!

Issue #144, opened 2023-04-19
rust:  Add `serde` feature to derive `Serialize` and `Deserialize` on all Protobuf definitions


romac:
This is needed for re-using the generated Rust proto definitions directly in the `ibc-proto` crate instead of duplicating them.

See https://github.com/cosmos/ibc-proto-rs/issues/10 and https://github.com/cosmos/ibc-rs/issues/640 for more details.

PR #145, opened 2023-04-19
feat: Add `serde` feature to derive `Serialize` and `Deserialize` on all Protobuf definitions

romac:
Closes: #144

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,8 +16,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  clippy:
-    name: Clippy
+  fmt:
+    name: Format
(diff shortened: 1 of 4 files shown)


romac:
CI failure is due to Codecov, I think we can just ignore it.

romac:
This should only require a patch version bump to 0.10.1 since adding trait impls for serde cannot cause compilation failures in downstream code afaik.

Alternatively, we could omit the derived serialization impls on the ibc.core.commitment protos in ibc-proto but that might break downstream code for those who rely on it.

Let me know what you think

romac:
Superseded by #146 which brings ProtoJSON-compatible `Serialize` and `Deserialize` instances, as per @hdevalence's suggestion.

PR #146, opened 2023-04-22
feat: Add ProtoJSON-compatible `Serialize` and `Deserialize` instances on all Protobuf definitions via `pbjson`

romac:
Closes: #144
Supersedes: #145


hdevalence (approved):
This would be great, and would unblock us on Penumbra.

tac0turtle (approved):

PR #153, opened 2023-06-19
feat: Add support for SMT `ExclusionProofs`

h5law:
This PR implements the support for SMT `ExclusionProofs` and verifies them by using converting them into an `ExistenceProof` and verifying either the existence of a `nil` value at the key provided or an unrelated key-value pair in place of the key provided.

_NOTE_: In the `LeafOp` of the `ExclusionProof` the reason we set the `PrehashKey` and `PrehashValue` to `HashOp_NO_HASH` is because the `ActualPath` and `ActualValueHash` fields should be populated by the `NonMembershipLeafData` field of the SMT proof and as such should already be hashed in accordance with the `TreeSpec` meaning hashing them again would result in an incorrect root computation.


h5law:
@crodriguezvega @AdityaSripal This is still a WIP and the following must be done:
- Add unit tests to the go library to cover the new type and any changes made
- Add support for Rust, and Javascript and add unit tests

h5law:
@crodriguezvega @AdityaSripal This PR is now ready for review. Go and rust implementations with test coverage are included. No JS support. This is partially due to me not seeing a usecase currently for JS and having no experience with TS. As it stands this logic will be used in our go project at Pocket and the rust will be used in my implementation of ICS-08 for our WASM client.

Makefile
```diff
@@ -1,5 +1,5 @@
 DOCKER := $(shell which docker)
-HTTPS_GIT := https://github.com/cosmos/cosmos-sdk.git
+HTTPS_GIT := https://github.com/cosmos/ics23.git
```
h5law, line 2: This change was made as the current `make proto-check-breaking` doesn't work it would say something about needing 2 contianers? I will revert this if someone can say how to make it work as intended.

go/proof.go
```diff
@@ -43,7 +43,7 @@ var TendermintSpec = &ProofSpec{
 	},
 }
 
-// SmtSpec constrains the format for SMT proofs (as implemented by github.com/celestiaorg/smt)
+// SmtSpec constrains the format for SMT proofs (as implemented by github.com/pokt-network/smt)
```
Olshansk, line 46: 🙌

go/proof.go
```diff
@@ -163,6 +168,95 @@ func (p *ExistenceProof) calculate(spec *ProofSpec) (CommitmentRoot, error) {
...
+func (p *ExclusionProof) Calculate() (CommitmentRoot, error) {
+	if p.GetLeaf() == nil {
+		return nil, errors.New("exclusion proof needs defined LeafOp")
+	}
+	if p.Leaf.PrehashKey != HashOp_NO_HASH {
+		return nil, errors.New("exclusion proof must have leaf with PrehashKey == NO_HASH")
+	}
+	if p.Leaf.PrehashValue != HashOp_NO_HASH {
+		return nil, errors.New("exclusion proof must have leaf with PrehashValue == NO_HASH")
+	}
```
Olshansk, line 180: Why is this a requirement? Consider adding a comment or updating the error message with the reasoning

go/proof.go
```diff
@@ -163,6 +168,95 @@ func (p *ExistenceProof) calculate(spec *ProofSpec) (CommitmentRoot, error) {
 	return res, nil
 }
```
Olshansk, line 170: I think that once we merge in your documentation in IBC, and even potentially add the mermaid diagram in [1], we should link to it here as a reference

[1] https://github.com/pokt-network/pocket/pull/845#discussion_r1245881510

go/proof.go
```diff
@@ -163,6 +168,95 @@ func (p *ExistenceProof) calculate(spec *ProofSpec) (CommitmentRoot, error) {
...
+	}
+
+	// leaf step takes the key and value as input
+	res, err := p.Leaf.Apply(p.ActualPath, p.ActualValueHash)
+	if err != nil {
+		return nil, fmt.Errorf("leaf, %w", err)
+	}
+	// check if the actual value was a placeholder and replace with empty hash
+	// assumes the placeholder value is [32]byte
+	if bytes.Equal(p.ActualValueHash, make([]byte, 32)) {
```
Olshansk, line 189: Readability NIT: Move `make([]byte, 32)` into a constant in this package or a local var and rename to `smtPlaceholder`

Olshansk (approved):
Left a couple minor NITS.

Approving because core business logic LGTM but requires review/approval from an owner of the repo as well.

crodriguezvega:
> @crodriguezvega @AdityaSripal This PR is now ready for review. Go and rust implementations with test coverage are included. No JS support. This is partially due to me not seeing a usecase currently for JS and having no experience with TS. As it stands this logic will be used in our go project at Pocket and the rust will be used in my implementation of ICS-08 for our WASM client.

Thank you, @h5law! We have also contacted the Penumbra folks and ask them to take a look at the PR.

Regarding JS support, we have talked in the past to actually stop support for it, so I wouldn't worry about it for now. If we end up still wanting to keep support, I could at some point port your Go implementation into TS.

Olshansk:
> > @crodriguezvega @AdityaSripal This PR is now ready for review. Go and rust implementations with test coverage are included. No JS support. This is partially due to me not seeing a usecase currently for JS and having no experience with TS. As it stands this logic will be used in our go project at Pocket and the rust will be used in my implementation of ICS-08 for our WASM client.
> 
> Thank you, @h5law! We have also contacted the Penumbra folks and ask them to take a look at the PR.
> 
> Regarding JS support, we have talked in the past to actually stop support for it, so I wouldn't worry about it for now. If we end up still wanting to keep support, I could at some point port your Go implementation into TS.

Would be great to connect with the Penumbra directly. In particular, I believe they might appreciate/adopt the work we're doing at https://github.com/pokt-network/smt

Olshansk:
@AdityaSripal Just wanted to follow up on what @h5law mentioned with a bit of a simplification and some diagrams (see [jmt]):

1. Empty node along path(key)
2. Different key at path(key)

Screenshot 2023-06-29 at 1 16 53 PM

Screenshot 2023-06-29 at 1 17 32 PM

[jmt](https://developers.diem.com/papers/jellyfish-merkle-tree/2021-01-14.pdf)

go/ics23.go
```diff
@@ -148,6 +179,23 @@ func getExistProofForKey(proof *CommitmentProof, key []byte) *ExistenceProof {
...
 
+func getExclusionProofForKey(proof *CommitmentProof, key []byte) *ExclusionProof {
+	switch p := proof.Proof.(type) {
+	case *CommitmentProof_Exclusion:
+		ep := p.Exclusion
+		if bytes.Equal(ep.Key, key) {
+			return ep
+		}
+	case *CommitmentProof_Batch:
+		for _, sub := range p.Batch.Entries {
```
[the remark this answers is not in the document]
h5law: We are checking the batch for the exclusion proof with the correct key. If there are multiple then they are the same proof and we need only return 1.

When we validate a batch in entirety we go through all proofs and execute them individually. This is just that we were given a batch entry and are looking to prove just the one with the matching key.

go/proof.go
h5law: I left a comment explaining this in the proto but will add again here

Olshansk:
I know there was a lengthy offline (slack/discord) discussion with @AdityaSripal @hdevalence and others related to not having `ExclusionProofs` in ics23, but rather just updating https://github.com/pokt-network/smt to support range proofs.  

@h5law could you capture and summarize the key takeaways and next steps?

cc @musalbas for context

h5law:
> @h5law could you capture and summarize the key takeaways and next steps?

The general sentiment is that instead of supporting newer proof mechanisms, the repo aims to support just a single form of `NonMembershipProof` using the "range based" (not exactly a range) technique already present. Although this is very specific to the IAVL tree, SMTs such as the JMT by penumbra have modified their implementation of the JMT to use this proof system instead of the intended `ExclusionProof` (proving the non-existence of a key via a sentinel value due to the sparse nature of the tree).

As a result of this intention it was suggested to have the SMT library instead support the same "range based" non-membership proof as the JMT and IAVL tree. As a result I have opened [this issue](https://github.com/pokt-network/smt/issues/16) in the SMT library.

I would like to add that although I will add support for this proof type I think that forcing newer tree types into using the same methods for generating proofs is not the way this library should be moving in. I feel that if we are able to support newer proof types, as long as the changes are backwards compatible we should do so. This enables the innovation to increase, potentially improvements in efficiency as time moves on. IBC is about interoperability not uniformity.

CC: @AdityaSripal @crodriguezvega @hdevalence @plaidfinch

jackzampolin:
Strongly agree with @h5law that only supporting a single proof method here is counter to the whole abstraction provided by ics23

faddat:
@jackzampolin -- I agree.  

There's the whole question of dragons and the like re: ics23, will dive in for a bit.

crodriguezvega:
Thanks for all the comments here, everyone. Our apologies to @h5law for not providing public feedback on the PR earlier. The reason for that is that discussions happened privately on Slack approximately a month ago, and the understanding of the people involved in them was that Harry would try Penumbra's suggestion of adding support for range-based non-inclusion proofs in Pocket's SMT implementation in the same way that Penumbra did for their JMT implementation, instead of adding support for a new type of exclusion proof to ics23. Even though Penumbra's and [Harry's views](https://github.com/cosmos/ics23/pull/153#issuecomment-1644286352) on how to support non-inclusion proofs for SMT differed, we had the impression that Harry would still try to implement Penumbra's suggestion. Unfortunately that was a misunderstanding from our side, and the consequence was that this PR did not get enough attention since then. Last week @jackzampolin brought this PR again to our attention, and conversations with Harry followed. We agreed that we will pick this up again in approximately one week from now, once @AdityaSripal is back from holidays.

Again, thanks everyone for the feedback and for putting this again in our radar. Whenever anybody notices that an issue of PR is falling through the cracks, feel free to ping us on Slack or Discord, or write a message on the relevant issue/PR. Due to our capacity constraints, we cannot treat everything with high priority, but with a bit of information and context, some rationale for prioritisation, and a pinch of kindness and understanding ❤️ we will do our best to make compromises to fit requests in as soon as possible.

crodriguezvega:
@h5law Did you have time to check out [Aditya's comment above](https://github.com/cosmos/ics23/pull/153#issuecomment-1708300457) and the [corresponding PR](https://github.com/cosmos/ics23/pull/186)?

h5law:
@crodriguezvega Yes sorry I forgot to update this PR, @AdityaSripal and I synced offline on this. I would like to pick this issue back up and have it in full feature parity to the SMT doing the check itself at a later date. I still believe this is an important feature add to the `ics23` repo and will build it in soon, once we have released the next iteration of our protocol.

The issue aditya pointed out is covered in the SMT itself but I looked over it here, although it took a long time Aditya's review will come in very helpful when I pick this up again in the near future.

As a result I will be closing this PR and will open a new one when the time comes.

PR #154, opened 2023-07-01
fix: Change `cosmos.ics23.v1.serde` mod to `no-std`.

riversyang:
## Motivation

During the integration of `ibc-rs` v0.41.0 (which uses `ics23` v0.10.1) into [near-ibc](https://github.com/octopus-network/near-ibc), I encountered compile errors in the `cosmos.ics23.v1.serde` module of `ics23`. The module was using several `std` types that are not allowed in NEAR smart contracts.

To resolve this, I made the necessary changes to allow compilation. I believe it would be beneficial to also commit these changes to the original repository.

## All Changes

* Replace `std` in `cosmos.ics23.v1.serde` to `core`.
* Add use for `format` macro and `Vec` of `alloc` in `cosmos.ics23.v1.serde`.

### DIFF
--- a/rust/src/cosmos.ics23.v1.serde.rs
+++ b/rust/src/cosmos.ics23.v1.serde.rs
@@ -1,6 +1,8 @@
+use alloc::{format, vec::Vec};
+
 impl serde::Serialize for BatchEntry {
     #[allow(deprecated)]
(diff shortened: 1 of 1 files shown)


romac:
Superseded by #158

PR #157, opened 2023-07-20
quick fix for no std

dzmitry-lahoda:
but really ibc-rs people should ensure it is usable in no_std envs like substrate/near/solana(yeah, solana can do IBC too:) ).

cw people  denying for year now idea of no_std support btw. sad.


DaviRain-Su:
cosmos.ics23.serde.rs is generate file by pbjson build. should fix on upstream.

romac:
Superseded by #158

PR #181, opened 2023-08-27
chore: avoid cloning and use `.map`/`.entry` in Rust code

kevinji:
There's no functional changes, just some code cleanup.


romac (approved):
Very nice, thanks!

PR #192, opened 2023-09-15
fix: fuzz test

crodriguezvega:
This should fix the error introduced [here](https://github.com/cosmos/ics23/pull/151).


go/fuzz_test.go
```diff
@@ -79,11 +79,11 @@ func FuzzVerifyNonMembership(f *testing.F) {
...
 	// 2. Now run the fuzzer.
 	f.Fuzz(func(t *testing.T, inputJSON []byte) {
-		bv := new(BatchVectorData)
-		if err := json.Unmarshal(inputJSON, bv); err != nil {
+		var bv BatchVectorData
+		if err := json.Unmarshal(inputJSON, &bv); err != nil {
 			return
 		}
-		if bv.Ref == nil || bv.Ref.RootHash == nil {
+		if bv.Ref == nil || bv.Proof == nil || bv.Ref.RootHash == nil {
```
crodriguezvega, line 86: The problem was that not checking if proof was `nil`.

colin-axner (approved):
ACK, not sure why the code cov is decreasing. Would also be nice to require passing go tests to merge prs

elias-orijtech:
LGTM. CC @odeke-em

odeke-em:
Thank you very much @crodriguezvega! Thank you for the tag @elias-orijtech

PR #202, opened 2023-09-28
chore(rust): update `prost` to v0.12

romac:


rust/codegen/Cargo.toml
```diff
@@ -8,6 +8,6 @@ edition = "2018"
 
 [dependencies]
 bytes = "1.0.1"
-prost = "0.11"
+prost = "0.12"
 prost-build = "0.12"
```
romac, line 12: For some reason, `prost-build` was already on v0.12.

hdevalence (approved):

hdevalence:
This counts as an API-breaking change for the Rust code, right?

What's the policy about crate versioning vs. ICS versioning? If we have a change to the Rust crate that's API-breaking but not different at the protocol level, how is that handled?

PR #205, opened 2023-10-03
chore(rust): bump version to 0.11.0

romac:
I had to yank v0.10.3 of the Rust crate because it was [breaking semver compat](https://github.com/cosmos/ics23/pull/202#issuecomment-1741255921) and breaking people's CI.

While we've so far tried to keep all versions in sync, I suggest we re-publish it as v0.11.0 and let the Go, TS and Rust versions drift. I added a compatibility table for Go and Rust in the README to let people know which versions of which implementation they can use. I am not sure exactly what the status is with the TS version, so I left it out for now. Help welcome.


damiannolan (approved):
Staggered versioning seems fine as long as the table is kept up to date! 🙏

PR #212, opened 2023-10-23
feat: Add support for Blake2b/2s/3 hash functions

larry0x:
This is a result of this Twitter discussion: https://twitter.com/hdevalence/status/1715765138547212689


README.md
```diff
@@ -6,7 +6,7 @@
 | ------------------ | ------------------------------------------------- | ---------------------------------------------- |
 | [Go](./go)         | [![Go Test][go-test-badge]][go-test-link]         | [![Go Cov][go-cov-badge]][go-cov-link]         |
 | [Rust](./rust)     | [![Rust Test][rust-test-badge]][rust-test-link]   | [![Rust Cov][rust-cov-badge]][rust-cov-link]   |
-| [TypeScript](./ts) | [![TypeScript Test][ts-test-badge]][ts-test-link] | [![TypeScript Cov][ts-cov-badge]][ts-cov-link] |
+| [TypeScript](./js) | [![TypeScript Test][ts-test-badge]][ts-test-link] | [![TypeScript Cov][ts-cov-badge]][ts-cov-link] |
```
larry0x, line 9: Fix a typo

go/Makefile
```diff
@@ -1,7 +1,7 @@
 test:
 	go test .
 
-.PHONY: test 
+.PHONY: test
```
larry0x, line 4: Remove a trailing whitespace

testdata/TestDoHashData.json
```diff
@@ -1,22 +1,42 @@
...
     "ExpectedHash": "b1ab9988c7c7c5ec4b2b291adfeeee10e77cdd46"
   },
-  "sha256": {
-    "HashOp": 1,
+  "bitcoin": {
+    "HashOp": 5,
     "Preimage": "food",
-    "ExpectedHash": "c1f026582fe6e8cb620d0c85a72fe421ddded756662a8ec00ed4c297ad10676b"
+    "ExpectedHash": "0bcb587dfb4fc10b36d57f2bba1878f139b75d24"
   },
```
larry0x, line 21: Prettifying this file by ordering test cases by the HashOp indexes. Also adding a test case for SHA2-512

rust/src/ops.rs
```diff
@@ -59,11 +62,26 @@ pub(crate) fn do_length(length: LengthOp, data: &[u8]) -> Result<Hash> {
...
+        LengthOp::Fixed64Big => {
+            let mut len = (data.len() as u64).to_be_bytes().to_vec();
+            len.extend(data);
+            return Ok(len);
+        }
+        LengthOp::Fixed64Little => {
+            let mut len = (data.len() as u64).to_le_bytes().to_vec();
+            len.extend(data);
+            return Ok(len);
+        }
```
larry0x, line 84: Implementations for these `LengthOp`s are missing, so I add them here because why not

go/ops.go
```diff
@@ -239,16 +259,28 @@ func doLengthOp(lengthOp LengthOp, data []byte) ([]byte, error) {
...
+	case LengthOp_FIXED64_LITTLE:
+		res := make([]byte, 8, 8+len(data))
+		binary.LittleEndian.PutUint64(res[:8], uint64(len(data)))
+		res = append(res, data...)
+		return res, nil
 		// TODO
 		// case LengthOp_VAR_RLP:
-		// case LengthOp_FIXED32_BIG:
-		// case LengthOp_FIXED64_BIG:
-		// case LengthOp_FIXED64_LITTLE:
```
larry0x, line 251: Add a few missing `LengthOp` impls

testdata/TestDoHashData.json
```diff
@@ -1,22 +1,42 @@
...
   "sha512_256": {
     "HashOp": 6,
     "Preimage": "food",
     "ExpectedHash": "5b3a452a6acbf1fc1e553a40c501585d5bd3cca176d562e0a0e19a3c43804e88"
+  },
+  "blakd2b": {
```
AdityaSripal, line 27:
```suggestion
  "blake2b": {
```

ditto for typos below

rust/src/proto_descriptor.bin
AdityaSripal, line 1: Remove

proto/cosmos/ics23/v1/proofs.proto
```diff
@@ -13,6 +13,9 @@ enum HashOp {
   RIPEMD160  = 4;
   BITCOIN    = 5; // ripemd160(sha256(x))
   SHA512_256 = 6;
+  BLAKE2B    = 7;
+  BLAKE2S    = 8;
```
AdityaSripal, line 17: Should we append with `_512` and `_256` to specify which variant we're using?

AdityaSripal (changes requested):
Thanks for this @larry0x! I think the proto names could be more specific.

Could you also add some more context to the PR description for future readers?

Specifically what is your use case for adding blake3? As you say, there's no official support in golang right now. This would mean that any chain using blake3 for its tree will not be able to connect to a cosmos-sdk chain, since we will use the go package of ICS23 that doesn't support this hash function. Seems strange to add a hash function that is incompatible with the vast majority of deployed ibc chains

rust/src/proto_descriptor.bin
larry0x: The Rust crate actually uses this file: https://github.com/cosmos/ics23/blob/rust/v0.11.0/rust/src/lib.rs#L21

larry0x:
> Specifically what is your use case for adding blake3? As you say, there's no official support in golang right now. This would mean that any chain using blake3 for its tree will not be able to connect to a cosmos-sdk chain, since we will use the go package of ICS23 that doesn't support this hash function. Seems strange to add a hash function that is incompatible with the vast majority of deployed ibc chains

@AdityaSripal This is true. There are two options: 1) use one of the _unofficial_ Go implementations, or 2) create a Wasm light client based on the Rust ICS-23. In fact I'm now leaning towards a Wasm client that uses a custom proof format, so whether this PR gets accepted will not impact my project. Your call.

testdata/TestDoHashData.json
```diff
@@ -1,22 +1,42 @@
...
+  },
+  "blakd2s": {
+    "HashOp": 8,
+    "Preimage": "food",
+    "ExpectedHash": "5a1ec796f11f3dfc7e8ca5de13828edf2e910eb7dd41caaac356a4acbefb1758"
+  },
+  "blakd3": {
+    "HashOp": 9,
+    "Preimage": "food",
+    "ExpectedHash": "f775a8ccf8cb78cd1c63ade4e9802de4ead836b36cea35242accf31d2c6a3697"
```
crodriguezvega, line 40: I will remove this because it's used in the go tests, but if there is no implementation for it yet, then the test will panic.

rust/src/ops.rs
crodriguezvega: I am also going to add tests for these and push to your branch. I hope that's ok with you.

crodriguezvega (commented):
Thank you for these additions, @larry0x! 

I am going to push now some tests for the extra length ops that you have implemented.

Regarding the support for blake3 in Go, there's someone we can contact for advice. Maybe one of the unofficial implementations is well positioned to eventually land in `x/crypto` and we can start already using it now.

crodriguezvega:
> @AdityaSripal This is true. There are two options: 1) use one of the unofficial Go implementations, or 2) create a Wasm light client based on the Rust ICS-23. In fact I'm now leaning towards a Wasm client that uses a custom proof format, so whether this PR gets accepted will not impact my project. Your call.

I messaged our Golang cryptography expert and this is what he says:

> There's little appetite for adding BLAKE3 to the standard library (x/crypto or otherwise) as of now, because it's not needed by anything else in the stdlib, it's not trivial to maintain, it needs to be heavily optimized (because why use BLAKE3 otherwise), and there are good 3rd party modules for it. I have not reviewed it extensively, but [lukechampine.com/blake3](http://lukechampine.com/blake3) is by a known author, looks well maintained, and its use of Avo, its tests, and its API are reassuring.

So we could the implementation that he recommends. What do you think, @AdityaSripal?

larry0x:
Hey guys, just want to clarify that I'm now leaning towards using my own custom proof format + Wasm light client. So, if you don't feel like supporting Blake in this implementation for any reason, it will not impact my project.

AdityaSripal:
I think i'd prefer not supporting blake3 in go until its official. I think requiring golang chains to use wasm light clients if they need to connect to a state tree that uses blake3 is acceptable.

@larry0x is there a reason you want to go with a custom proof format over ICS23 beyond the fact this isn't released yet?

larry0x:
> is there a reason you want to go with a custom proof format over ICS23 beyond the fact this isn't released yet?

For my case a custom proof format can be more compact and probably cheaper to run, especially when it comes to non-existent proofs, because my tree is a radix tree, I only need to show that a node doesn't have a certain children, I don't need to show that two siblings both exist as is how ICS23 works. @AdityaSripal

romac:
The Rust code looks good to me. Thanks for adding tests @crodriguezvega! 🙏

testdata/TestDoHashData.json
```diff
@@ -1,22 +1,37 @@
...
+    "HashOp": 7,
+    "Preimage": "food",
+    "ExpectedHash": "b1f115361afc179415d93d4f58dc2fc7d8fa434192d7cb9b65fca592f6aa904103d1f12b28655c2355478e10908ab002c418dc52a4367d8e645309cd25e3a504"
+  },
+  "blake2s": {
+    "HashOp": 8,
+    "Preimage": "food",
+    "ExpectedHash": "5a1ec796f11f3dfc7e8ca5de13828edf2e910eb7dd41caaac356a4acbefb1758"
   }
 }
```
AdityaSripal, line 37: no blake3 test?

AdityaSripal: Ahh right, thanks

Issue #211, opened 2023-10-23
Clarify which specific hash algorithm `HashOp::Keccak` refers to


larry0x:
It's unclear whether it refers to Keccak-256, -512, or other variants.

There is no Go or TypeScript implementation of this HashOp. The Rust implementation uses Sha3-512, which is distinct from either  Keccak-256 or -512. None of the implementations has a test case for this.

Would appreciate if this confusion is addressed.

crodriguezvega:
Thank you for opening this issue, @larry0x. This is a very good catch.

May I ask if you're planning to use a variant of Keccak for your use case then?

We should try to find if anybody is using the current `HashOp_Keccak` enum value. If nobody is, then I guess we could:

- Update the name of the enum value to be more explicit of what variant it is used.
- Update the Rust code to use the correct hash operation instead of SHA3-512.
- Optionally add also support in Go and TS code.

But the first two changes are API breaking, so we need to be careful and figure out if we can make those without disruption.

Also I would like to hear @romac's thoughts on this.

romac:
These changes are fine by me, since we do not use Keccak in Hermes, nor in ibc-rs as far as I can tell. Let me page @Farhad-Shabani just to be sure, and to also run these changes by him.

Farhad-Shabani:
> nor in ibc-rs as far as I can tell.

That's right. I also reached out to a potential user, and they are not using the Keccak variant as well. So we should be all set.

PR #217, opened 2023-11-02
fix: clarify `HashOp::Keccak` variant to refer to Keccak-256

Farhad-Shabani:
Closes: #211

### Description
Considering that no use cases have been reported for the current SHA-3 Rust implementation of the `HashOp::Keccak`, and with the goal of bolstering support for meaningful use cases, I have opted for the inclusion of Keccak-256 as the most suitable variant (Ethereum uses this particular hash function).

### Note
For the JS part, I couldn't properly generate code through existing commands, so it made me to adjust things there manually. That would be great if @webmaster128 could look into it.


webmaster128:
> For the JS part, I couldn't properly generate code through existing commands, so it made me to adjust things there manually. That would be great if @webmaster128 could look into it.

Diff looks sane as far as I can tell. I'd not invest time in the current proto codegen for JS anymore. We should replace that with Telescope code generation.

rust/src/host_functions.rs
```diff
@@ -48,9 +48,9 @@ pub mod host_functions_impl {
             buf
         }
 
-        fn sha3_512(message: &[u8]) -> [u8; 64] {
```
crodriguezvega, line 51: This is being removed because it's not used?

Farhad-Shabani: That’s right. As far as I could check in projects around `ibc-rs`, they are not using it, and there isn't any `HashOp` variant already for it.

hdevalence:
> Considering that no use cases have been reported for the current SHA-3 Rust implementation of the HashOp::Keccak, and with the goal of bolstering support for meaningful use cases, I have opted for the inclusion of Keccak-256 as the most suitable variant (Ethereum uses this particular hash function).

Wait, but SHA-3 is the standard one. It differs from Keccak-256 only in some minor constants. Wouldn't it be better to have support for both standard SHA-3 and the Ethereum-specific variant, considering that there's effectively no extra code required?

Farhad-Shabani:
> Wait, but SHA-3 is the standard one. It differs from Keccak-256 only in some minor constants. Wouldn't it be better to have support for both standard SHA-3 and the Ethereum-specific variant, considering that there's effectively no extra code required?

I'm cool with adding a standard option. However, just to be clear, the SHA-3 implementation we've had so far is SHA3-512 (64 bytes). Afaik, the one that's more in line with Keccak256 is SHA3-256.
Honestly, I haven't come across any major projects using SHA3-512. Maybe we could focus on nailing down the Keccak variant first here and then tackle supporting SHA3 as a separate effort after some evaluation and seeing which algorithm covers the most relevant use cases. What do you think?

crodriguezvega (approved):
Thank you, @Farhad-Shabani. Sorry it took a while to review. 

The changes look good to me, I just added a [test](https://github.com/cosmos/ics23/pull/217/commits/3ad31165332d2360c3f6e4b2115d4adb53611ac5).

Farhad-Shabani:
> Thank you, @Farhad-Shabani. Sorry it took a while to review.

Thank you so much @crodriguezvega for reviewing this :pray:

romac (approved):
LGTM

PR #268, opened 2024-01-18
fix: trying to fix error in JS code coverage upload

crodriguezvega:


DimitrisJim (approved):
lgtm. this codecov failure seems out of funk, previous diff to this didn't add/remove code/test code.

PR #286, opened 2024-02-13
ci: pass in codecov secret for go, typescript.

DimitrisJim:
Hopefully will take care of these failures on the workflows.


crodriguezvega (approved):
Yay, nice one. Thank you for fixing this, @DimitrisJim!

Let's see if we can figure out the next problem as well: why it complains that the code coverage decreseaes... 😓

PR #291, opened 2024-03-08
ci: add sonarcloud for go and js

crodriguezvega:


crodriguezvega:
> seems like this was triggered in this PR too [#290 (comment)](https://github.com/cosmos/ics23/pull/290#issuecomment-1985587040) that doesn't touch go/typescript.
> 
> we might need to tweak the `sonar.exclusions`?

Yeah, I am not sure if we can prevent that, because the project in SonarCloud was created for the whole repository... I didn't add a `sonar-project.properties` file for Rust, because it looks like [code coverage support for Rust is not supported natively yet](https://richardwbown.com/code-analysis-in-sonarcloud-using-github-actions/)?

But I will update `sonar.exclusions` anyway.

PR #306, opened 2024-04-17
feat: add api folder containing pulsar generated code.

DimitrisJim:
Required for https://github.com/cosmos/ibc-go/pull/5968#

Protos contain only primitive types so generation was straigh-forward. Would be a good idea to add a step to the go workflow that enters the mod and does a sanity `go build ./...`


damiannolan:
Thanks @DimitrisJim! 

Happy to leave this open or switch to draft. Likely won't need to merge this if we don't need protov2 and can use gogoproto.

damiannolan:
This can be closed as pulsar is no longer required. see https://github.com/cosmos/cosmos-sdk/issues/20134

PR #307, opened 2024-04-26
docs: add documentation

crodriguezvega:
For rendered view, see [this](https://github.com/cosmos/ics23/blob/efd149657c10934bce7f92307651a15c25bb071a/docs/README.md).

Opened followup issue #385

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 # ICS 23
 
-As already stated in the README.md, ics23 attempts to be a generic library to represent and verify merkle proofs for (ideally) many different merkle tree storage implementations. The library is heavily used in [ibc-go](https://github.com/cosmos/ibc-go) for proof verification. The following documentation uses the Golang implementation as a reference.
+As already stated in the README.md, ICS 23 attempts to be a generic library to represent and verify merkle proofs for (ideally) many different merkle tree storage implementations. The library is heavily used in [ibc-go](https://github.com/cosmos/ibc-go) for proof verification. The following documentation uses the Golang implementation as a reference.
 
 The two most important top level types in the library are `ProofSpec` and `CommitmmentProof`. We will explain them in the coming sections.
 


docs/README.md
```diff
@@ -0,0 +1,225 @@
...
+  HashOp         hash              = 6;
+}
+```
+
+where:
+
+- `child_order` is the ordering of the children node and counts from 0. For example, for an IAVL tree is [0, 1] (left, then right); for a merk tree is [0, 2, 1] (left, right, here).
+- `child_size`
+- `min_prefix_length`
+- `max_prefix_length`
```
crodriguezvega, line 66: @AdityaSripal If you could explain me a bit what these field are exactly for, that would be great. I have a bit of hard time understanding these ones.

docs/README.md
```diff
@@ -0,0 +1,225 @@
+# ics23
```
colin-axner, line 1:
```suggestion
# ICS 23
```

docs/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+As already stated in the README.md, ics23 attempts to be a generic library to represent and verify merkle proofs for (ideally) many different merkle tree storage implementations. The library is heavily used in [ibc-go](https://github.com/cosmos/ibc-go) for proof verification. The following documentation uses the Golang implementation as a reference.
```
colin-axner, line 3:
```suggestion
As already stated in the README.md, ICS 23 attempts to be a generic library to represent and verify merkle proofs for (ideally) many different merkle tree storage implementations. The library is heavily used in [ibc-go](https://github.com/cosmos/ibc-go) for proof verification. The following documentation uses the Golang implementation as a reference.
```

colin-axner (approved):
Reviewed the documentation additions. They look good to me. I think the debugger images are a bit noisy, but I figure it's better to have something than nothing

Issue #348, opened 2024-08-08
Add test tree tooling


colin-axner:
We should be able to generate test cases using some test tree structure rather than manually inputting leaf and inner node values

PR #352, opened 2024-08-14
fix: apply default restrictions to max depth

colin-axner:
If the max depth is set to 0, treat it as 128


testdata/TestCheckAgainstSpecData.json
```diff
@@ -7,16 +7,16 @@
         "hash": 1,
```
crodriguezvega, line 7: Curious: how did you re-generate this test data?

crodriguezvega (commented):
Thank you for the fix, @colin-axner. Should we add a changelog entry?

testdata/TestCheckAgainstSpecData.json
```diff
@@ -259,113 +259,670 @@
...
+        "child_order": [
+          0,
+          1
+        ],
+        "child_size": 33,
+        "min_prefix_length": 4,
+        "max_prefix_length": 12,
+        "hash": 1
+      },
+      "max_depth": 4,
```
crodriguezvega, line 786: Shouldn't actually this be 0 so that it uses the default value of 128?

```suggestion
      "max_depth": 0,
```

crodriguezvega (commented):
Should we also update [this comment](https://github.com/cosmos/ics23/blob/472162105c71b0c37b2f4c13dc2f4892bad2a499/proto/cosmos/ics23/v1/proofs.proto#L165) on the proto file?

It looks like we should also fix this in the [Rust implementation](https://github.com/cosmos/ics23/blob/472162105c71b0c37b2f4c13dc2f4892bad2a499/rust/src/verify.rs#L122-L133), right? (I can try to open a PR or contact @romac otherwise)

go/proof_data_test.go
```diff
@@ -54,7 +54,7 @@ func CheckLeafTestData(tb testing.TB) map[string]CheckLeafTestStruct {
 type CheckAgainstSpecTestStruct struct {
 	Proof *ExistenceProof
 	Spec  *ProofSpec
-	IsErr bool
+	Err   string
```
damiannolan, line 57: should we make this `error` instead of `string`?

damiannolan: maybe not worth doing this tbh. I see the testdata json structs are unmarshalled to this struct and we'd need to handle custom unmarshalJSON functionality to handle the built-in error interface. 

Using `string` is fine by me 👍🏻

go/proof.go
```diff
@@ -190,7 +190,13 @@ func (p *ExistenceProof) CheckAgainstSpec(spec *ProofSpec) error {
 	if spec.MinDepth > 0 && len(p.Path) < int(spec.MinDepth) {
 		return fmt.Errorf("innerOps depth too short: %d", len(p.Path))
 	}
-	if spec.MaxDepth > 0 && len(p.Path) > int(spec.MaxDepth) {
+
+	maxDepth := spec.MaxDepth
+	if maxDepth == 0 {
+		maxDepth = 128
+	}
```
damiannolan, line 197: Should we add a note to proto doc that if `max_depth` is zero then we treat as default 128?

https://github.com/cosmos/ics23/blob/master/proto/cosmos/ics23/v1/proofs.proto#L165-L166

testdata/TestCheckAgainstSpecData.json
damiannolan: +1, were they changed by hand?

damiannolan (approved):
Thanks @colin-axner! LGTM

go/proof_data_test.go
colin-axner: yup, I tried doing `error` but felt it was too much work to dig further into

testdata/TestCheckAgainstSpecData.json
colin-axner: yup! The only way I managed was to pull printed proofs from ibc-go tests :sweat_smile: That's why I added #348

testdata/TestCheckAgainstSpecData.json
colin-axner: ah yes! thank you

go/proof.go
colin-axner: Sure thing :+1:

testdata/TestCheckAgainstSpecData.json
damiannolan: Ah yea, figured as much when I saw the issue! 😅 sometimes caveman approach = best approach

AdityaSripal (approved):
Thanks for the great work @colin-axner !!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+# Unreleased
+
+- fix(go): interpret max_depth in inner specs as 128 if left to 0 [#352](https://github.com/cosmos/ics23/pull/352).
```
colin-axner, line 5:
```suggestion
- fix(go): interpret max_depth in proof specs as 128 if left to 0 [#352](https://github.com/cosmos/ics23/pull/352).
```

rust/src/proto_descriptor.bin
colin-axner, line 1: this was added when I ran `make proto-all`. Hope it is correct if I update it @romac

PR #353, opened 2024-08-15
fix: remove typescript implementation

crodriguezvega:
Remove typescript implementation, since there is no known usage of it and it is not actively maintained.


romac (approved):

Gldywn:
fwiw if anyone still needs JS/TS support, I’ve built [@gldywn/ics23.js](https://github.com/Gldywn/ics23.js), wasm bindings of the `ics23` Rust crate with a clean Node.js API. It supports membership / non-membership verification, batch proofs, compression/decompression, default specs and root calculation.

If maintainers are open to it, I'd be glad to open a PR to add a link in the README 👍

PR #360, opened 2024-08-29
fix: a misnamed rust function

### LINKED ISSUE #324: Rust: misnamed hash function
In [host_functions.rs](https://github.com/cosmos/ics23/blob/master/rust/src/host_functions.rs#L54-L59):

```rust
pub trait HostFunctionsProvider {
    /// The SHA-512 hash algorithm with its output truncated to 256 bits.
    fn sha2_512_truncated(message: &[u8]) -> [u8; 32];
}
```

Then, in [ops.rs](https://github.com/cosmos/ics23/blob/master/rust/src/ops.rs#L48), this function is used to do the SHA512/256 hash:

```rust
pub(crate) fn do_hash<H: HostFunctionsProvider>(hash: HashOp, data: &[u8]) -> Hash {
    match hash {
        HashOp::Sha512256 => Hash::from(H::sha2_512_truncated(data)),
        // ...
    }
}
```

This is incorrect: "SHA2-512 truncated" and SHA512/256 are two distinct things:

- "SHA2-512 truncated" means to do the SHA2-512 hash, which produces a 64-byte hash, and manually truncate off the second half;
- SHA512/256 is a different hash algorithm (also part of the SHA2 family) that natively produces a 32-byte hash.

According to the proto definition, the intention is to use SHA512/256, meaning the function `sha2_512_truncated` is misnamed. It should be renamed to `sha2_512_256`. However, the implementation is correct.

Issue caught by @Rhaki

larry0x:
Closes: #324

See #324 for rationale.

### DIFF
--- a/rust/src/host_functions.rs
+++ b/rust/src/host_functions.rs
@@ -8,8 +8,8 @@ pub trait HostFunctionsProvider {
     /// The SHA-512 hash algorithm
     fn sha2_512(message: &[u8]) -> [u8; 64];
 
-    /// The SHA-512 hash algorithm with its output truncated to 256 bits.
-    fn sha2_512_truncated(message: &[u8]) -> [u8; 32];
+    /// The SHA2-512/256 hash algorithm.
+    fn sha2_512_256(message: &[u8]) -> [u8; 32];
 
     /// The Keccak-256 hash function.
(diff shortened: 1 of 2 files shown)


damiannolan (approved):
Looks correct to me, would be nice to add a changelog.

PR #364, opened 2024-08-29
fix: return error if branch not found in child order

crodriguezvega:


go/proof.go
```diff
@@ -373,82 +416,94 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
 	}
 	for i, item := range order {
 		if branch == item {
-			return i
+			return i, nil
 		}
 	}
-	panic(fmt.Errorf("branch %d not found in order %v", branch, order))
+
+	return 0, fmt.Errorf("branch %d not found in order %v", branch, order)
```
crodriguezvega, line 495: This is basically the change needed. The rest of changes are for bubbling the error up.

go/proof.go
```diff
@@ -373,82 +416,94 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
+	return true, nil
 }
 
 // getPosition checks where the branch is in the order and returns
 // the index of this branch
-func getPosition(order []int32, branch int32) int {
+func getPosition(order []int32, branch int32) (int, error) {
 	if branch < 0 || int(branch) >= len(order) {
-		panic(fmt.Errorf("invalid branch: %d", branch))
+		return 0, fmt.Errorf("invalid branch: %d", branch)
```
crodriguezvega, line 487: I also converted this one into a regular error instead of panicking.

go/proof.go
crodriguezvega: Alternatively I could have just swallowed the error in the call sites of `getPosition` and return in those places the default value (false, and 0s). I thought of bubbling the error up in case we ever do https://github.com/cosmos/ics23/issues/29.

go/proof.go
```diff
@@ -373,82 +416,94 @@ func getPadding(spec *InnerSpec, branch int32) (minPrefix, maxPrefix, suffix int
...
 }
 
 // leftBranchesAreEmpty returns true if the padding bytes correspond to all empty siblings
 // on the left side of a branch, ie. it's a valid placeholder on a leftmost path
-func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp) bool {
+func leftBranchesAreEmpty(spec *InnerSpec, op *InnerOp) (bool, error) {
 	idx, err := orderFromPadding(spec, op)
 	if err != nil {
-		return false
+		return false, err
```
crodriguezvega, line 427: Here an error that previously was swallowed can now be returned.

go/proof.go
```diff
@@ -277,30 +292,46 @@ func (p *NonExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []b
...
 	for _, step := range path {
-		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !leftBranchesAreEmpty(spec, step) {
-			return false
+		leftBranchesAreEmpty, err := leftBranchesAreEmpty(spec, step)
+		if err != nil {
+			return false, err
+		}
+
+		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !leftBranchesAreEmpty {
+			return false, nil
```
colin-axner, line 309: todo: did this short circuit before if `hasPadding` returned true? (ie would that have an effect if we now call left branches first?

go/proof.go
```diff
@@ -277,30 +292,46 @@ func (p *NonExistenceProof) Verify(spec *ProofSpec, root CommitmentRoot, key []b
...
 	for _, step := range path {
-		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !rightBranchesAreEmpty(spec, step) {
-			return false
+		rightBranchesAreEmpty, err := rightBranchesAreEmpty(spec, step)
+		if err != nil {
+			return false, err
+		}
+
+		if !hasPadding(step, minPrefix, maxPrefix, suffix) && !rightBranchesAreEmpty {
+			return false, nil
```
colin-axner, line 331: ditto

colin-axner (commented):
todo: add test case to cover change from panic to error in `getPosition`

go/proof.go
colin-axner: I decided to change the code such that it would execute with the exact same code trace as before:
@crodriguezvega changed it to:
```go
leftBranchesAreEmpty, err := leftBranchesAreEmpty(spec, step)
if err != nil {
	return false, err
}

if !hasPadding(step, minPrefix, maxPrefix, suffix) && !leftBranchesAreEmpty {
	return false, nil
}
```

I changed it to:
```go
if !hasPadding(step, minPrefix, maxPrefix, suffix) {
        leftBranchesAreEmpty, err := leftBranchesAreEmpty(spec, step)
        if err != nil {
            return false, err 
        }   

        if !leftBranchesAreEmpty {
            return false, nil 
        }   
}
```

Because golang has short circuiting, `!hasPadding` would have previously resulted in `rightBranchesAreEmpty()` not being called, but in the code @crodriguezvega refactored, it would have been called on every execution. I tried to look at the correctness of such change, if `rightBranchAreEmpty` would function regardless of the determined padding, but it was too exhausting of a mental exercise to justify the minor differences in readability, so I opt'd to leave execution as it was before

cc @AdityaSripal

AdityaSripal (approved):

colin-axner (approved):
reviewed again myself

PR #365, opened 2024-08-29
fix: prevent out of bounds on lookup inners access

crodriguezvega:
When decompressing a compressed proof. If the compressed proof has a path value (index) which is out of bounds of the inner op lookup list, return an error (previously would panic)


go/compress.go
```diff
@@ -153,7 +153,9 @@ func decompressExist(exist *CompressedExistenceProof, lookup []*InnerOp) *Existe
 		Path:  make([]*InnerOp, len(exist.Path)),
 	}
 	for i, step := range exist.Path {
-		res.Path[i] = lookup[step]
+		if int(step) < len(lookup) {
```
AdityaSripal, line 156: Why are we just ignorign this case rather than returning nil or error?

AdityaSripal (commented):
General question on if it makes sense to just skip if we are out of bounds or return some error

go/compress.go
crodriguezvega: I don't know what it's best. I choose this, but I could equally do something like:

```go
for i, step := range exist.Path {
  if int(step) >= len(lookup) {
    // return error or panic?
  }
  res.Path[i] = lookup[step]
}
```

What do you think should be the best approach?

go/compress.go
```diff
@@ -106,45 +112,62 @@ func compressStep(step *InnerOp, lookup *[]*InnerOp, registry map[string]int32)
...
 		},
-	}
+	}, nil
 }
 
-func decompressExist(exist *CompressedExistenceProof, lookup []*InnerOp) *ExistenceProof {
+func decompressExist(exist *CompressedExistenceProof, lookup []*InnerOp) (*ExistenceProof, error) {
 	if exist == nil {
-		return nil
+		return nil, nil
```
colin-axner, line 170: tests broke with an error return here. I don't have a great understanding of when this case arises? Is there a comment we can add to elaborate on it? @AdityaSripal

AdityaSripal: If a nonexistence proof is proving the leftmost key wouldn't the leftExist be nil? So decompressExist returns nil and we don't error out on line 149

colin-axner: ah yes, thank you! That makes sense to me, added a comment

AdityaSripal:
Is this pr still needed if we're removing batch proofs?

colin-axner:
> Is this pr still needed if we're removing batch proofs?

Nope, changes are no longer relevant!

PR #369, opened 2024-09-10
fix: guarantee that max prefix length is < min prefix length + child size

crodriguezvega:


AdityaSripal (approved):
Nice work! Ideally we would also have a test that malleates the prefix in order to fool the proof system that now fails. You should be able to pull out the test case from the audit report so don't have to create yourself

rust/src/verify.rs
```diff
@@ -612,19 +623,32 @@ mod tests {
...
-                            valid_inner,
+                            valid_inner.clone(),
                         ],
                     },
                     spec: depth_limited_spec,
                     valid: false,
                 },
             ),
+            (
+                "rejects inner spec with max prefix length >= min prefix lenght + child size",
```
colin-axner, line 640:
```suggestion
                "rejects inner spec with max prefix length >= min prefix length + child size",
```

colin-axner (approved):
Agree that we should have the malleate test case (also a unit test for go code?). The audit recommended updating the proto documentation to indicate this restriction as well

colin-axner:
requesting review from @romac to verify the rust changes

colin-axner:
gonna mark as draft until:
- [ ] malleate test added
- [ ] unit test for check against inner spec
- [ ] proto docs updated
- [ ] rust formatting fixed

romac:
> requesting review from @romac to verify the rust changes

LGTM

go/ops.go
```diff
@@ -166,6 +166,10 @@ func (op *InnerOp) CheckAgainstSpec(spec *ProofSpec, b int) error {
 		return errors.New("spec.InnerSpec.ChildSize must be >= 1")
 	}
 
+	if spec.InnerSpec.MaxPrefixLength >= spec.InnerSpec.MinPrefixLength+spec.InnerSpec.ChildSize {
```
colin-axner, line 169: for what it is worth. Spec fields being validated against themselves are better off in a spec.Validate() function, but for the sake of minimizing changes, this approach makes sense to me

damiannolan (approved):

PR #373, opened 2024-09-25
test: add unit tests for go InnerOp.CheckAgainstSpec

colin-axner:


go/ops_test.go
```diff
@@ -61,3 +64,160 @@ func TestDoHash(t *testing.T) {
...
+			func() {
+				spec.InnerSpec.ChildSize = 0
+			},
+			errors.New("spec.InnerSpec.ChildSize must be >= 1"),
+		},
+		{
+			"failure: inner op suffix malformed",
+			func() {
+				// change spec to be non-iavl spec to skip strict iavl validation
+				spec.LeafSpec.PrehashKey = HashOp_BITCOIN
```
damiannolan, line 157: Don't think we need to change the hashOp in this testcase here, right?

Is this just for consistency with other testcases where prefix is modified?

damiannolan (approved):
Awesome! Thanks @colin-axner!

go/ops_test.go
colin-axner: You are correct! Nice catch! I have removed it now. Ideally that hashOp change wouldn't be in the test cases, but to save some time I decided to use that trick for a few of the more difficult test cases

PR #382, opened 2024-10-08
chore: remove file filter on proto workflow

colin-axner:
run buf breaking on every pr (takes about 10s)


colin-axner:
I mad the protobuf break check required as well @damiannolan

damiannolan (approved):
Nice, thank you!

PR #383, opened 2024-10-10
refactor: deobfuscate dragonberry

colin-axner:
Thank you @damiannolan for the assistance!

### DIFF
--- a/go/ops.go
+++ b/go/ops.go
@@ -20,38 +20,56 @@ import (
 	_ "golang.org/x/crypto/ripemd160" //nolint:staticcheck
 )
 
-// validate the IAVL Ops
-func validateIavlOps(op opType, b int) error {
+// validateIavlOps validates the prefix to ensure it begins with
+// the height, size, and version of the IAVL tree. Each varint must be a bounded value.
+// In addition, the remaining bytes are validated to ensure they correspond to the correct
+// length.
+func validateIavlOps(op opType, layerNum int) error {
 	r := bytes.NewReader(op.GetPrefix())
 
-	values := []int64{}
-	for i := 0; i < 3; i++ {
-		varInt, err := binary.ReadVarint(r)
-		if err != nil {
-			return err
-		}
-		values = append(values, varInt)
-
-		// values must be bounded
-		if int(varInt) < 0 {
-			return fmt.Errorf("wrong value in IAVL leaf op")
-		}
+	height, err := binary.ReadVarint(r)
+	if err != nil {
+		return err
 	}
-	if int(values[0]) < b {
-		return fmt.Errorf("wrong value in IAVL leaf op")
+	if int(height) < 0 || int(height) < layerNum {
+		return fmt.Errorf("IAVL height (%d) must be non-negative and less than the layer number (%d)", height, layerNum)
 	}
 
-	r2 := r.Len()
-	if b == 0 {
-		if r2 != 0 {
-			return fmt.Errorf("invalid op")
+	size, err := binary.ReadVarint(r)
+	if err != nil {
+		return err
+	}
+
+	if int(size) < 0 {
+		return fmt.Errorf("IAVL size must be non-negative")
+	}
+
+	version, err := binary.ReadVarint(r)
+	if err != nil {
+		return err
+	}
+
+	if int(version) < 0 {
+		return fmt.Errorf("IAVL version must be non-negative")
+	}
+
+	// grab the length of the remainder of the prefix
+	remLen := r.Len()
+	if layerNum == 0 {
+		if remLen != 0 {
+			return fmt.Errorf("expected remaining prefix to be 0, got: %d", r2)
 		}
 	} else {
-		if !(r2^(0xff&0x01) == 0 || r2 == (0xde+int('v'))/10) {
+		// when the child comes from the left, the suffix if filled in
+		// prefix: height | size | version | length byte (1 remainder)
+		//
+		// when the child comes from the right, the suffix is empty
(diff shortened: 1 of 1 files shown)


go/ops.go
```diff
@@ -20,38 +20,56 @@ import (
...
-func validateIavlOps(op opType, b int) error {
+// validateIavlOps validates the prefix to ensure it begins with
+// the height, size, and version of the IAVL tree. Each varint must be a bounded value.
+// In addition, the remaining bytes are validated to ensure they correspond to the correct
+// length.
+func validateIavlOps(op opType, layerNum int) error {
 	r := bytes.NewReader(op.GetPrefix())
 
-	values := []int64{}
-	for i := 0; i < 3; i++ {
```
colin-axner, line 28: I decided to remove the for loop for readability, so each height, size, and version could have a var name rather than an index

go/ops.go
```diff
@@ -20,38 +20,56 @@ import (
...
-		varInt, err := binary.ReadVarint(r)
-		if err != nil {
-			return err
-		}
-		values = append(values, varInt)
+	height, err := binary.ReadVarint(r)
+	if err != nil {
+		return err
+	}
+	if int(height) < 0 || int(height) < layerNum {
```
colin-axner, line 34: previous check: `if int(values[0]) < b {` moved here in `int(height) < layerNum`

go/ops.go
```diff
@@ -20,38 +20,56 @@ import (
 	_ "golang.org/x/crypto/ripemd160" //nolint:staticcheck
 )
 
-// validate the IAVL Ops
-func validateIavlOps(op opType, b int) error {
+// validateIavlOps validates the prefix to ensure it begins with
+// the height, size, and version of the IAVL tree. Each varint must be a bounded value.
+// In addition, the remaining bytes are validated to ensure they correspond to the correct
+// length.
+func validateIavlOps(op opType, layerNum int) error {
```
colin-axner, line 27: `b` -> `layerNum` open to other naming suggestions

go/ops.go
```diff
@@ -20,38 +20,56 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	if int(version) < 0 {
+		return fmt.Errorf("IAVL version must be non-negative")
+	}
+
+	// grab the length of the remainder of the prefix
+	remLen := r.Len()
```
colin-axner, line 57: `r2` -> `remLen` open to other naming suggestions!

go/ops.go
```diff
@@ -20,38 +20,56 @@ import (
...
+	if layerNum == 0 {
+		if remLen != 0 {
+			return fmt.Errorf("expected remaining prefix to be 0, got: %d", remLen)
```
colin-axner, line 60:
```suggestion
			return fmt.Errorf("expected remaining prefix length to be 0, got: %d", remLen)
```

go/ops.go
damiannolan: `layerNum` is good to me, maybe we can indicate in godoc that it corresponds to tree depth

colin-axner:
cc @valardragon (if you want to take a peek :slightly_smiling_face: )

go/ops.go
```diff
@@ -20,38 +20,56 @@ import (
 	_ "golang.org/x/crypto/ripemd160" //nolint:staticcheck
 )
 
-// validate the IAVL Ops
-func validateIavlOps(op opType, b int) error {
+// validateIavlOps validates the prefix to ensure it begins with
+// the height, size, and version of the IAVL tree. Each varint must be a bounded value.
+// In addition, the remaining bytes are validated to ensure they correspond to the correct
+// length. The layerNum is the tree depth.
```
damiannolan, line 26: added two more lines on leaf node/inner node constraints (feel free to take or leave) 

```suggestion
// validateIavlOps validates the prefix to ensure it begins with
// the height, size, and version of the IAVL tree. Each varint must be a bounded value.
// In addition, the remaining bytes are validated to ensure they correspond to the correct
// length. The layerNum is the tree depth. 
// For leaf nodes (depth=0), we ensure the op prefix is constrained to the height, size and
// version of the IAVL tree.
// For inner nodes (depth>=1), we ensure the remaining prefix length conforms to the expected
// length when applying inner nodes where child comes from either left or right (see comment inline).
```

damiannolan (approved):
Beautiful, thank you @colin-axner! So much easier to grok! ❤️

go/ops.go
damiannolan: on reflection this reads a little weird. `layerNum` is passed as 0 when we're at the leaf, but in reality that is depth n, since depth of 0 usually means root. 

maybe we could change to:

```suggestion
// validateIavlOps validates the prefix to ensure it begins with
// the height, size, and version of the IAVL tree. Each varint must be a bounded value.
// In addition, the remaining bytes are validated to ensure they correspond to the correct
// length. The layerNum is the inverse of the tree depth, i.e. depth=0 means leaf, depth>=1 means inner node
```

colin-axner: I like your second suggestion. It's concise and with the additional in-code comment, I think it's straightforward enough to reason about

faddat:
love it

go/ops_test.go
```diff
@@ -9,6 +9,124 @@ import (
...
+		},
+		{
+			"failure: invalid remaining length with layer number is 0",
+			func() {
+				layerNum = 0
+			},
+			fmt.Errorf("expected remaining prefix length to be 0, got: 1"),
+		},
+		{
+			"failure: invalid reminaing length with non-zero layer number",
```
AdityaSripal, line 89:
```suggestion
			"failure: invalid remaining length with non-zero layer number",
```

PR #386, opened 2024-10-17
feat: add stricter checks for iavl spec and tendermint spec

colin-axner:


go/ops.go
```diff
@@ -75,6 +84,26 @@ func validateIavlOps(op opType, layerNum int) error {
 	return nil
 }
 
+// validateTendermintOps validates the prefix to ensure it begins with []byte{1}.
+func validateTendermintOps(op *InnerOp) error {
```
colin-axner, line 88: for [reference](https://github.com/cosmos/cosmos-sdk/blob/f01baf302e2bbe9648cf454372115dfd1488b663/store/internal/proofs/convert.go#L56)

damiannolan (approved):
🙏🏻

AdityaSripal (approved):

PR #419, opened 2026-05-13
docs: harmless second workflow approval probe for release/v0.10.x

qsq8t65pxs:
Minimal harmless follow-up probe to verify whether a second external PR from the same outsider account still requires workflow approval before untrusted pull_request workflows execute. No exploit payload, no workflow edits, no secret-touching changes.


qsq8t65pxs:
Closing this minimal harmless workflow-approval probe now that the approval-gating behavior has been confirmed.

